PR #60, opened 2022-07-02
Fix rojo not found for sourcemaps on macOS

filiptibell:
Spawning the child process for creating the Rojo sourcemap now has a fallback where the shell configured in vscode settings is spawned interactively and the `rojo sourcemap` command runs in there. This means that the shell is responsible for loading the environment and `PATH` variable instead of vscode / the daemon that starts up GUI apps for macOS.

### DIFF
--- a/editors/code/src/spawn.ts
+++ b/editors/code/src/spawn.ts
@@ -1,135 +1,133 @@
-import * as vscode from 'vscode';
+import * as vscode from "vscode";
 
 import {
-	ChildProcessWithoutNullStreams,
-	SpawnOptionsWithoutStdio,
-	spawn,
-} from 'child_process';
+  ChildProcessWithoutNullStreams,
+  SpawnOptionsWithoutStdio,
+  spawn,
+} from "child_process";
 
 type ChildProcessRejection = {
-	interactive: boolean,
-	code: number,
-	reason: string,
+  interactive: boolean;
+  code: number;
+  reason: string;
 };
 
-
-
-
-
 const canSpawnInteractive = (platform: NodeJS.Platform): boolean => {
-	switch (platform) {
-		case 'darwin':
-			return true;
-		case 'linux':
-			return true;
-		default:
-			return false;
-	}
+  switch (platform) {
+    case "darwin":
+      return true;
+    case "linux":
+      return true;
+    default:
+      return false;
+  }
 };
 
 const handleChildProcessPromise = (
-	cp: ChildProcessWithoutNullStreams,
-	resolve: (value: string) => void,
-	reject: (reason: ChildProcessRejection) => void,
-	interactive: boolean,
+  cp: ChildProcessWithoutNullStreams,
+  resolve: (value: string) => void,
+  reject: (reason: ChildProcessRejection) => void,
+  interactive: boolean
 ) => {
-	// Set stdout and stderr encoding + pipe to plain strings
-	let stdout = "";
-	let stderr = "";
-	cp.stdout.setEncoding('utf8');
-	cp.stderr.setEncoding('utf8');
-	cp.stdout.on("data", data => stdout += data);
-	cp.stderr.on("data", data => stderr += data);
-	// Reject if spawning errors out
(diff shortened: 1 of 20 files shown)


editors/code/src/spawn.ts
```diff
@@ -0,0 +1,133 @@
...
+
+const spawnNaive = (
+  command: string,
+  args: string[],
+  options?: SpawnOptionsWithoutStdio
+): Promise<string> => {
+  return new Promise((resolve, reject) => {
+    // Spawn the wanted command directly
+    const cp = spawn(command, args, {
+      cwd: process.cwd(),
```
JohnnyMorganz, line 70: cwd should use the one passed in options if available
```suggestion
      cwd: options?.cwd ?? process.cwd(),
```

editors/code/src/spawn.ts
```diff
@@ -0,0 +1,133 @@
...
+  return new Promise((resolve, reject) => {
+    // Spawn the configured vscode shell as if it is a
+    // normal interactive shell + set the proper env
+    // vars needed for an interactive shell
+    let cwd = process.cwd();
+    if (options && typeof options.cwd === "string") {
+      cwd = options.cwd;
+    }
+    const cp = spawn(vscode.env.shell, {
+      cwd: cwd,
```
JohnnyMorganz, line 94: I think this cwd can be simplified similar to above?
```suggestion
    const cp = spawn(vscode.env.shell, {
      cwd: options?.cwd ?? process.cwd(),
```

editors/code/src/spawn.ts
```diff
@@ -0,0 +1,133 @@
...
+    spawnNaive(command, args, options)
+      .then(resolve)
+      .catch((err) => {
+        if (!err.ran && canSpawnInteractive(process.platform)) {
+          spawnInteractive(command, args, options).then(resolve).catch(reject);
+        } else {
+          reject(err);
+        }
+      });
+  });
```
JohnnyMorganz, line 130: I don't think this needs to be wrapped inside another promise
```suggestion
  return spawnNaive(command, args, options)
      .catch((err) => {
        if (!err.ran && canSpawnInteractive(process.platform)) {
          return spawnInteractive(command, args, options);
        } else {
          return Promise.reject(err);
        }
      });
```

editors/code/src/spawn.ts
```diff
@@ -0,0 +1,133 @@
...
+const wrappedSpawn = (
+  command: string,
+  args: string[],
+  options?: SpawnOptionsWithoutStdio
+): Promise<string> => {
+  return new Promise((resolve, reject) => {
+    spawnNaive(command, args, options)
+      .then(resolve)
+      .catch((err) => {
+        if (!err.ran && canSpawnInteractive(process.platform)) {
```
JohnnyMorganz, line 124: Where does `err.ran` come from? I can't seem to find it

JohnnyMorganz (changes requested):

editors/code/src/spawn.ts
filiptibell: `cwd` should be overwritten by any options passed in with the `...options` on line `72`

editors/code/src/spawn.ts
filiptibell: For this one I pulled out `cwd` to be able to pass it as the `PWD` env var on line `97`, is there a more elegant way to do that?

editors/code/src/spawn.ts
filiptibell: Leftover from some earlier testing, removed it!

editors/code/src/spawn.ts
JohnnyMorganz: Ah yeah, makes sense, my bad

editors/code/src/spawn.ts
JohnnyMorganz: Whoops, didn't see that. I don't think so

JohnnyMorganz (approved):
Thanks!

Issue #29, opened 2022-06-11
Register extra definition files depending on file


JohnnyMorganz:
It would be good to be able to register extra definitions, e.g. `testez.d.lua`.

We should put this in configuration and make it so it maps globs to extra definitions

mikejohnstn:
@JohnnyMorganz I'm not sure how much work this issue will be to implement properly, so in the meantime, would it be possible to just expose this file after the extension is installed?
https://github.com/JohnnyMorganz/luau-lsp/blob/master/scripts/globalTypes.d.lua

If we could simply add to or replace this file (in the VS Code extensions directory) post-installation, that'd be perfect for now.

JohnnyMorganz:
@mikejohnstn You should be able to find the file installed at `%appdata%\Code\User\globalStorage\johnnymorganz.luau-lsp` (or the non-Windows equivalent). Note that it does automatically run an update every Roblox update (which is weekly), so you may have to rewrite it every so often.

Hopefully I'll get something to sort this out soon (and probably another option to disable the auto API update if you don't need it!)

mikejohnstn:
Found the file, and works perfectly for now. Thanks!

mikejohnstn:
I updated and confirmed the new definition file is working, as well as disabling the Roblox definition file. 👍 

A couple issues (and maybe why you haven't closed this yet):

1) If there's an error while reading the definition file, you only get a very generic message. Is there a way to surface what the problem might be? Otherwise it takes a lot of guessing when you have a typo or other problem in the definitions. My workaround is to manually reload after pretty much every small change to my definition file to make sure it works.

2) Any way to auto-refresh when the definition file changes, so we don't have to manually reload?

JohnnyMorganz:
Definitely want to improve both of these aspects, the issue right now is that definition files are passed and loaded at server startup (using command line flags). We can't display diagnostics until the initialisation handshake is complete, but we load in the file before this. Loading at startup is also why a reboot is required currently.

I plan on looking into deferring the definition files until later, just need to find out a way to cleanly clear the old types and reload the new ones, which is the main hurdle right now. Forcing a reload right now is just me being lazy 😅

> 1. If there's an error while reading the definition file, you only get a very generic message. Is there a way to surface what the problem might be? Otherwise it takes a lot of guessing when you have a typo or other problem in the definitions. My workaround is to manually reload after pretty much every small change to my definition file to make sure it works.

[...]

PR #84, opened 2022-07-23
Add environment support

JohnnyMorganz:
Adds support for type environments.

Environments can now be configured using `luau-lsp.types.environments` and `luau-lsp.types.environmentGlobs`.
You can create an environment name and point it to a definitions file, then assign an environment to a glob.

This is useful e.g. by creating a "testez" environment, only applicable to "*.spec.lua". TestEZ related global types will then only be available in those particular files, and will not pollute the global types.

In future, the environment name will be used for documentation parsing.

This extends onwards from #29 

`luau-lsp.types.definitionFiles` has been deprecated in favour of environments. Assign the glob "*" to a global environment.

Blocked on https://github.com/Roblox/luau/pull/610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Deprecated
+
+- `luau-lsp.types.definitionFiles` has been deprecated in favour of environment support. Instead, create an environment in `luau-lsp.types.environments` and then assign a glob "\*" to your environment.
+
+### Added
+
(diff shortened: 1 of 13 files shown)


JohnnyMorganz:
This feature is currently blocked on the following:

> We modify `typeCheckerForAutocomplete` to provide more expressive DM types not present in the base type checker (which AIUI Studio does as well). For environment scopes though, there is no differentiation between the two so we can't do this. Environment scopes only build off the base typeChecker global scope

m4fh:
Is this still blocked? I'm making a Rojo-like tool that uses Luau and this would be extremely useful.

JohnnyMorganz:
I believe this feature is blocked until the new type solver becomes standard

PR #108, opened 2022-08-18
Correct WaitForChild type

### LINKED ISSUE #106: Incorrect WaitForChild return type
It looks like the `WaitForChild` method returns type `Instance` even when given a timeout which would cause it to return `nil` instead.

Qualadore:
Closes #106.

### DIFF
--- a/scripts/dumpRobloxTypes.py
+++ b/scripts/dumpRobloxTypes.py
@@ -155,7 +155,8 @@ EXTRA_MEMBERS = {
         "function FindFirstChild(self, name: string, recursive: boolean?): Instance?",
         "function FindFirstDescendant(self, name: string): Instance?",
         "function GetActor(self): Actor?",
-        "WaitForChild: ((self, name: string) -> Instance) & ((self, name: string, timeout: number) -> Instance?)",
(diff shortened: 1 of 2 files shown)


scripts/dumpRobloxTypes.py
```diff
@@ -154,6 +155,7 @@
         "function FindFirstChild(self, name: string, recursive: boolean?): Instance?",
         "function FindFirstDescendant(self, name: string): Instance?",
         "function GetActor(self): Actor?",
+        "WaitForChild: ((self, name: string) -> Instance) & ((self, name: string, timeout: number) -> Instance?)",
```
JohnnyMorganz, line 158: The recommended way that I would do this is by using the class function syntax like the rest of them. If you define more than one with the same name, Luau automatically converts it into an overloaded function
```suggestion
        "function WaitForChild(self, name: string): Instance",
        "function WaitForChild(self, name: string, timeout: number): Instance?"
```

JohnnyMorganz (changes requested):

JohnnyMorganz (approved):
Thanks!

PR #118, opened 2022-09-03
Add RBXScriptSignal:Once

Baileyeatspizza:
Attempts to fix 
image

PR #122, opened 2022-09-04
Revert "Improve RequestAsync type"

JohnnyMorganz:
Reverts JohnnyMorganz/luau-lsp#121


JohnnyMorganz:
Looks like this wasn't the issue, it was #118

PR #125, opened 2022-09-06
Correct types for `HumanoidDescription` and `TeleportService` methods

fewkz:
Adds some type corrections for `HumanoidDescription` and `TeleportService` methods
- `GetPlayerPlaceInstanceAsync`'s type was incorrect and any proper usage of it would cause issues.
- Many `TeleportService` methods took `{ Instance }` for players, rather than `{ Player }`
- `TeleportService:TeleportAsync` took `Instance` instead of `TeleportOptions` for `teleportOptions` parameter
- Added types for `HumanoidDescription`'s methods that used to just take `any` 
- `TeleportService:ReserveServer`'s type returned `any`, instead of `(string, string)`, so it would return `*error-type*` for the second value


JohnnyMorganz (approved):
Thank you!

PR #128, opened 2022-09-06
Correct types for `TeleportService` events

fewkz:
These used `Instance` rather than `Player`, and in `TeleportInitFailed`, used `Instance` instead of `TeleportOptions`


JohnnyMorganz (approved):
Thanks!

PR #130, opened 2022-09-12
Correct type for `UserService:GetUserInfosByUserIdsAsync`

fewkz:
Replaces the definition of `function GetUserInfosByUserIdsAsync(self, userIds: { any }): { any }` with `	function GetUserInfosByUserIdsAsync(self, userIds: { number }): { { Id: number, Username: string, DisplayName: string } }`


JohnnyMorganz (approved):
Thanks!

PR #134, opened 2022-09-14
Correct GuiObject tweening types

Qualadore:
Callback parameters weren't being typed as optional.


JohnnyMorganz:
The underlying cause is because we don't wrap the callback in parentheses before appending `?`
https://github.com/JohnnyMorganz/luau-lsp/blob/b1fe1ccf03e1255c2cb11254bcf5966853d69577/scripts/globalTypes.d.lua#L5840
Instead we should just always create all callbacks wrapped in parens

JohnnyMorganz:
Closed in favour of #135

PR #150, opened 2022-09-24
BasePart type improvements

fewkz:
Many methods used `Instance` rather than `BasePart`
`GetNetworkOwner` may be nil


JohnnyMorganz (approved):
Thanks!

PR #153, opened 2022-09-25
More type improvements

fewkz:
Fixes types:
`Camera.CameraSubject` could possibly be `nil`. It's valid types are `BasePart` and `Humanoid`
`Team:GetPlayers()` return a list of `Player`s
`Teams:GetTeams` return a list of `Team`s
`AncestryChanged`'s second signal argument may be nil if the ancestor's parent became `nil`.
`FindFirstChild...` and `FindFirstAncestor...` methods didn't return `Instance?`
`WorldRoot:GetPart...` returns an array of `{ BasePart }`
and some more

`GetPartsObscuringTarget` does in fact take a list of `Instance`s, not `BasePart`s

`Teams:GetTeams()` causes the type definition file to fail to load due to complexity, so the tests will probably fail. I can split it off from this PR.

### DIFF
--- a/scripts/dumpRobloxTypes.py
+++ b/scripts/dumpRobloxTypes.py
@@ -342,7 +342,7 @@ EXTRA_MEMBERS = {
         "function GetPlayers(self): { Player }"
     ],
     "Camera": [
-        "CameraSubject: Instance | BasePart | nil",
+        "CameraSubject: Humanoid | BasePart | nil",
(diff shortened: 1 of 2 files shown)


JohnnyMorganz:
Need to regen with the new changes

fewkz:
wait dont merge

fewkz:
done 👍

PR #158, opened 2022-09-26
Correct return type for Teams:GetTeams()

fewkz:
This will probably fail because it causes too much complexity in the type definitions file


fewkz:
oh huh, it passed this time

PR #165, opened 2022-09-27
Fix duplicate methods for plugins

fewkz:
Overridden members weren't listed in IGNORED_MEMBERS

### DIFF
--- a/scripts/globalTypes.d.lua
+++ b/scripts/globalTypes.d.lua
@@ -7927,7 +7927,6 @@ declare class PluginPolicyService extends Instance
 end
 
 declare class PluginToolbar extends Instance
-	function CreateButton(self, buttonId: string, tooltip: string, iconname: string, text: string?): PluginToolbarButton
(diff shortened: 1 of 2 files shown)


fewkz:
actually, It looks like the entry for `Plugin:CreateToolbar()` could be removed entirely

scripts/globalTypes.d.lua
```diff
@@ -7927,7 +7927,6 @@ declare class PluginPolicyService extends Instance
 end
 
 declare class PluginToolbar extends Instance
-	function CreateButton(self, buttonId: string, tooltip: string, iconname: string, text: string?): PluginToolbarButton
 	function CreateButton(self, id: string, toolTip: string, iconAsset: string, text: string?): PluginToolbarButton
```
JohnnyMorganz, line 7930: Do we even need this CreateButton override as well?

fewkz: I think the parameter names is an improvement, so may as well keep it
`iconAsset` is clearer that it's an image rather than `iconname`, when I first saw `iconname` in the tooltip I was confused what it did

PR #170, opened 2022-09-29
Plugin QoL changes

Barocena:
- Shortcut Binding for Connect/Disconnect 
- PortTextBox only accepting numbers with 5 character limit now
- PortTextBox don't clear on focus now (to prevent accidently clicking and clearing textbox)
- Changed the Name of Toolbar button (since it just opens the menu and to not confuse with connect binding on shortcut settings)
- Error handling on empty port
- Fixed scaling of Luau Icon

### DIFF
--- a/plugin/src/init.server.lua
+++ b/plugin/src/init.server.lua
@@ -3,8 +3,12 @@ local HttpService = game:GetService("HttpService")
 assert(plugin, "This code must run inside of a plugin")
 
 local toolbar = plugin:CreateToolbar("Luau") :: PluginToolbar
-local button =
-	toolbar:CreateButton("Language Server Setup", "Toggle Menu", "rbxassetid://11115506617", "Luau LSP") :: PluginToolbarButton
+local button = toolbar:CreateButton(
+	"Luau Language Server Setup",
+	"Toggle Menu",
+	"rbxassetid://11115506617",
+	"Luau Language Server"
+) :: PluginToolbarButton
 
 local widgetInfo = DockWidgetPluginGuiInfo.new(
 	-- widget info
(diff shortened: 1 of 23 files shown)


plugin/src/init.server.lua
```diff
@@ -194,3 +195,14 @@ end)
...
+
+ConnectAction.Triggered:Connect(function()
+	if connected then
+		print("[Luau Language Server] Disconnecting from DataModel changes")
+		cleanup()
+	else
+		print("[Luau Language Server] Listening for DataModel changes")
+		watchChanges()
+	end
+	connectButton.Text = if connected then "Disconnect" else "Connect"
```
JohnnyMorganz, line 207: Can we extract this out into a common function used by both button and action?

plugin/src/init.server.lua
```diff
@@ -3,12 +3,8 @@ local HttpService = game:GetService("HttpService")
...
 
 local toolbar = plugin:CreateToolbar("Luau") :: PluginToolbar
-local button = toolbar:CreateButton(
-	"Language Server Setup",
-	"Connect to Server",
-	"rbxassetid://10913079454",
-	"Connect to Server"
-) :: PluginToolbarButton
+local button =
+	toolbar:CreateButton("Language Server Setup", "Toggle Menu", "rbxassetid://11115506617", "Luau LSP") :: PluginToolbarButton
```
JohnnyMorganz, line 7:
```suggestion
	toolbar:CreateButton("Language Server Setup", "Toggle Menu", "rbxassetid://11115506617", "Luau Language Server") :: PluginToolbarButton
```

Can you change references of LSP to Language Server (I made a mistake of "Luau LSP" as it is technically wrong, but it's difficult to change existing references elsewhere)

JohnnyMorganz (commented):
Thanks!

Barocena:
Fixed the naming and created common function for connecting to server 
also changed prints logs a bit to make it clear (it was printing connected before giving error)
and added small indicator for easier use with shortcuts only

image

JohnnyMorganz (approved):
LGTM

Issue #176, opened 2022-10-08
Ignore glob patterns for standalone analyze


filiptibell:
It is currently a bit tricky to get the standalone `luau-lsp analyze` command working with the same behavior as the VSCode extension, since it also emits errors for files that have been required, and not just the files given to the command. Ignore glob patterns would solve this in the same way that the extension does.

PR #197, opened 2022-10-17
Add ignore glob support for analyze

metrowaii:
Adds support for ignoring errors from certain files using the glob pattern. The option is prefixed with; `--ignoreGlobs=`, adding multiple options with the same prefix should also work. 

This MR addresses [#176 ](https://github.com/JohnnyMorganz/luau-lsp/issues/176)

Note: I'm fairly "new" to C++ had some university experience with it, but I'm certainly not a pro. With that in mind, please criticize anything that I could fix/improve!

### DIFF
--- a/src/AnalyzeCli.cpp
+++ b/src/AnalyzeCli.cpp
@@ -50,9 +50,30 @@ static void report(ReportFormat format, const char* name, const Luau::Location& 
     }
 }
 
-static void reportError(const Luau::Frontend& frontend, ReportFormat format, const Luau::TypeError& error)
+static bool isIgnoredFile(const std::filesystem::path& rootUriPath, const std::filesystem::path& path, std::vector<std::string>& ignoreGlobPatterns) {
+    auto relativePath = path.lexically_relative(rootUriPath).generic_string(); // HACK: we convert to generic string so we get '/' separators
+
+    //luau analyze returns relative path for files that are to be analyzed
+    if (relativePath.empty())
+        relativePath = path.generic_string();
+
+    for (auto& pattern : ignoreGlobPatterns)
+        if (glob::fnmatch_case(relativePath, pattern))
+            return true;
+
+    return false;
+}
+
+static void reportError(const Luau::Frontend& frontend, ReportFormat format, const Luau::TypeError& error, std::vector<std::string>& ignoreGlobPatterns)
 {
-    std::string humanReadableName = frontend.fileResolver->getHumanReadableModuleName(error.moduleName);
+    WorkspaceFileResolver* fileResolver = static_cast<WorkspaceFileResolver*>(frontend.fileResolver);
+    std::filesystem::path rootUriPath = fileResolver->rootUri.fsPath();
+    std::string humanReadableName = fileResolver->getHumanReadableModuleName(error.moduleName);
+    auto path = fileResolver->resolveToRealPath(error.moduleName);
+
+    if (isIgnoredFile(rootUriPath, *path, ignoreGlobPatterns)) {
+        return;
+    }
 
     if (const Luau::SyntaxError* syntaxError = Luau::get_if<Luau::SyntaxError>(&error.data))
         report(format, humanReadableName.c_str(), error.location, "SyntaxError", syntaxError->message.c_str());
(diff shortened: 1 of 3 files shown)


src/AnalyzeCli.cpp
```diff
@@ -80,8 +95,19 @@ static bool analyzeFile(Luau::Frontend& frontend, const char* name, ReportFormat
         return false;
     }
 
-    for (auto& error : cr.errors)
+    WorkspaceFileResolver* fileResolver = static_cast<WorkspaceFileResolver*>(frontend.fileResolver);
```
JohnnyMorganz, line 98: I don't think we need to static_cast here

src/AnalyzeCli.cpp
```diff
@@ -80,8 +95,19 @@ static bool analyzeFile(Luau::Frontend& frontend, const char* name, ReportFormat
...
+    std::filesystem::path rootUriPath = fileResolver->rootUri.fsPath();
+
+    for (auto& error : cr.errors) 
+    {
+        auto path = fileResolver->resolveToRealPath(error.moduleName);
+
+        if (isIgnoredFile(rootUriPath, *path, ignoreGlobPatterns)) {
+            continue;
+        }
```
JohnnyMorganz, line 107: Could we move the `isIgnoredFile` check into `reportError`? `fileResolver` is already passed to it, so you can still get the rootUriPath

src/AnalyzeCli.cpp
```diff
@@ -67,7 +68,21 @@ static void reportWarning(ReportFormat format, const char* name, const Luau::Lin
     report(format, name, warning.location, Luau::LintWarning::getName(warning.code), warning.text.c_str());
 }
 
-static bool analyzeFile(Luau::Frontend& frontend, const char* name, ReportFormat format, bool annotate)
+static bool isIgnoredFile(const std::filesystem::path& rootUriPath, const std::filesystem::path& path, std::vector<std::string>& ignoreGlobPatterns) {
+    auto relativePath = path.lexically_relative(rootUriPath).generic_string(); // HACK: we convert to generic string so we get '/' separators
+
+    //luau analyze returns relative path for files that are to be analyzed
+    if (relativePath.empty())
+        relativePath = path;
```
JohnnyMorganz, line 76: When will this get hit? I can't think of when relativePath would be empty

JohnnyMorganz (commented):
Thanks! Just a few small things.

It would also be neat if we had tests for this, but we don't have any test harness yet for CLI tests - that would probably be a bit involved

src/AnalyzeCli.cpp
JohnnyMorganz: Actually nevermind, we probably do, since `frontend.fileResolver` doesn't know its a WorkspaceFileResolver

src/AnalyzeCli.cpp
metrowaii: This gets hit when the original path is already relative to the root path. The path returned by the file resolver is relative to the root path for the files that are passed into the analyze command, however, for files that contain errors outside of the files that you have specified in the command, the path is absolute. This check was initially introduced to handle the case where the file you're trying to analyze is also a file you're trying to ignore (e.g `luau-lsp analyze src/shared/test.lua --ignoreGlobs=src/shared/test.lua`). In this case, the relative path conversion returned an empty string since the path was already relative to the root path. Which I'm not sure is a case we should be solving for in the first place, but I could imagine a situation where someone might want to see lints, but not errors. Thoughts?

JohnnyMorganz: That inconsistency is somewhat weird (files passed are relative but the rest are not).
I think this is fine for now

src/AnalyzeCli.cpp
```diff
@@ -134,6 +157,8 @@ int startAnalyze(int argc, char** argv)
             // Backwards compatibility
             else if (strncmp(argv[i], "--defs=", 7) == 0)
                 definitionsPaths.push_back(std::string(argv[i] + 7));
+            else if (strncmp(argv[i], "--ignoreGlobs=", 14) == 0)
+                ignoreGlobPatterns.push_back(std::string(argv[i] + 14));
```
JohnnyMorganz, line 161: Probably better off just naming this `--ignore`
```suggestion
            else if (strncmp(argv[i], "--ignore=", 9) == 0)
                ignoreGlobPatterns.push_back(std::string(argv[i] + 9));
```

src/main.cpp
```diff
@@ -46,6 +46,7 @@ static void displayHelp(const char* argv0)
     printf("  --timetrace: record compiler time tracing information into trace.json\n");
     printf("  --sourcemap=PATH: path to a Rojo-style sourcemap\n");
     printf("  --definitions=PATH: path to definition file for global types\n");
+    printf("  --ignoreGlobs=GLOB: path(s) to ignore error outputs\n");
```
JohnnyMorganz, line 49:
```suggestion
    printf("  --ignore=GLOB: glob pattern to ignore error outputs\n");
```

src/AnalyzeCli.cpp
```diff
@@ -80,8 +102,8 @@ static bool analyzeFile(Luau::Frontend& frontend, const char* name, ReportFormat
         return false;
     }
 
-    for (auto& error : cr.errors)
-        reportError(frontend, format, error);
+    for (auto& error : cr.errors)
```
JohnnyMorganz, line 105:
```suggestion
    for (auto& error : cr.errors)
```

JohnnyMorganz (approved):
Looks good to me, thanks!

PR #222, opened 2022-11-02
Add FloatCurveKey and RotationCurveKey type

RealEthanPlayzDev:
I noticed there was no type of ``FloatCurveKey`` (https://robloxapi.github.io/ref/type/FloatCurveKey.html) and ``RotationCurveKey`` (https://robloxapi.github.io/ref/type/RotationCurveKey.html), so I've made a PR to add the types for those two.

Hopefully I didn't make a mistake here while editing the file

### DIFF
--- a/scripts/dumpRobloxTypes.py
+++ b/scripts/dumpRobloxTypes.py
@@ -393,8 +393,6 @@ type ProtectedString = string
 type BinaryString = string
 type QDir = string
 type QFont = string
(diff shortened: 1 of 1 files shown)


RealEthanPlayzDev:
Resolved in the following RobloxLsp pr: https://github.com/NightrainsRbx/RobloxLsp/pull/191
Should be there when auto-generation happens

PR #260, opened 2023-01-07
Support autocomplete function callbacks as arguments

### LINKED ISSUE #244: Autocomplete anonymous functions when passing as argument
If we need to pass an argument as a function (e.g. a callback), it would be nice if it there is a snippet option to autocomplete the parameters and go directly into the function body

JohnnyMorganz:
Closes #244 

We had to edit the `RBXScriptSignal` type so that it included parameter names, which is where this is most useful

TODO: this change currently breaks signature help because the label detection is no longer correct. We need better label detection -> maybe we need to port our own ToStringNamedFunction?

https://user-images.githubusercontent.com/19635171/211158546-ab849b12-f2ec-4a43-987f-e5a78694e079.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Added syntax highlighting support for interpolated strings
 - Added color viewers for Color3.new/fromRGB/fromHSV/fromHex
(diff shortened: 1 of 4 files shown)


JohnnyMorganz:
This feature is now actually built into Luau as "function (anonymous autofilled)" autocompletion.

We may need to improve the ergonomics a little, but it works for now.

PR #303, opened 2023-02-21
Fix Handles/ArcHandles.Adornee type

TreehouseFalcon:
Fix the `Handles.Adornee` & `ArcHandles.Adornee` property type (the property can be nil).
Additionally, fix the `Highlight.Adornee` property from a previous change I made. The Adornee property should probably only stick to BaseParts per the Roblox docs.

### DIFF
--- a/scripts/globalTypes.d.lua
+++ b/scripts/globalTypes.d.lua
@@ -5902,14 +5902,13 @@ declare class SelectionSphere extends PVAdornment
 end
 
 declare class PartAdornment extends GuiBase3d
-	Adornee: BasePart
+	Adornee: BasePart?
 end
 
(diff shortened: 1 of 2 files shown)


scripts/globalTypes.d.lua
```diff
@@ -5909,6 +5909,7 @@ declare class HandlesBase extends PartAdornment
 end
 
 declare class ArcHandles extends HandlesBase
+	Adornee: BasePart?
```
JohnnyMorganz, line 5912: I think this change needs to be made in a parent class. Adornee is inherited from PartAdornment

JohnnyMorganz (changes requested):

JohnnyMorganz (approved):

PR #311, opened 2023-03-07
Implement Auto-Import of requires

HawDevelopment:
Add auto-import for requires. Only works with virtual paths.

https://user-images.githubusercontent.com/70876593/223535126-c1b23b84-218d-400d-ab8e-3c05f93247be.mp4

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -44,15 +44,6 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
     return std::nullopt;
 }
 
-size_t pathDistance(const std::string& path1, const std::string& path2)
-{
-    std::filesystem::path p1(path1);
-    std::filesystem::path p2(path2);
-
-    auto distance = std::filesystem::relative(p1, p2);
-    return std::distance(distance.begin(), distance.end()) - 1;
-}
-
 std::string convertToScriptPath(const std::string& path)
 {
     std::filesystem::path p(path);
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -44,6 +44,43 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
     return std::nullopt;
 }
 
+size_t pathDistance(const std::string& path1, const std::string& path2)
+{
+    std::filesystem::path p1(path1);
+    std::filesystem::path p2(path2);
+
+    auto distance = std::filesystem::relative(p1, p2);
+    return std::distance(distance.begin(), distance.end()) - 1;
+}
+
+std::string convertToScriptPath(const std::string& path)
+{
+    std::filesystem::path p(path);
+    std::string output = "";
+    for (auto it = p.begin(); it != p.end(); ++it)
+    {
+        auto str = it->string();
+        if (str.find(" ") != std::string::npos)
+            output += "[\"" + str + "\"]";
+        else if (str == ".") {
+            if (it == p.begin())
+                output += "script";
+        }
+        else if (str == "..") {
+            if (it == p.begin())
+                output += "script.Parent";
+            else
+                output += ".Parent";
+        }
+        else {
+            if (it != p.begin()) output += ".";
+            output += str;
+        }
+    }
+    return output;
+}
+
+
 std::string codeBlock(const std::string& language, const std::string& code)
 {
     return "```" + language + "\n" + code + "\n" + "```";
--- a/src/include/LSP/ClientConfiguration.hpp
+++ b/src/include/LSP/ClientConfiguration.hpp
@@ -90,12 +90,10 @@ struct ClientCompletionConfiguration
     bool addTabstopAfterParentheses = true;
     /// If parentheses are added, fill call arguments with parameter names
     bool fillCallArguments = true;
-    /// The max 'path distance' before an auto-import require is relative
-    size_t maxRequirePathDistance = 4;
 };
 
 NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(
-    ClientCompletionConfiguration, enabled, suggestImports, addParentheses, addTabstopAfterParentheses, fillCallArguments, maxRequirePathDistance);
+    ClientCompletionConfiguration, enabled, suggestImports, addParentheses, addTabstopAfterParentheses, fillCallArguments);
 
 struct ClientSignatureHelpConfiguration
 {
--- a/src/include/LSP/ClientConfiguration.hpp
+++ b/src/include/LSP/ClientConfiguration.hpp
@@ -90,10 +90,12 @@ struct ClientCompletionConfiguration
     bool addTabstopAfterParentheses = true;
     /// If parentheses are added, fill call arguments with parameter names
     bool fillCallArguments = true;
+    /// The max 'path distance' before an auto-import require is relative
+    size_t maxRequirePathDistance = 4;
 };
 
 NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(
-    ClientCompletionConfiguration, enabled, suggestImports, addParentheses, addTabstopAfterParentheses, fillCallArguments);
+    ClientCompletionConfiguration, enabled, suggestImports, addParentheses, addTabstopAfterParentheses, fillCallArguments, maxRequirePathDistance);
 
 struct ClientSignatureHelpConfiguration
 {
--- a/src/include/LSP/Utils.hpp
+++ b/src/include/LSP/Utils.hpp
@@ -10,6 +10,8 @@
 
 std::optional<std::string> getParentPath(const std::string& path);
 std::optional<std::string> getAncestorPath(const std::string& path, const std::string& ancestorName);
+size_t pathDistance(const std::string& path1, const std::string& path2);
+std::string convertToScriptPath(const std::string& path);
 std::string codeBlock(const std::string& language, const std::string& code);
 std::optional<std::string> readFile(const std::filesystem::path& filePath);
 void trim_start(std::string& str);
--- a/src/include/LSP/Utils.hpp
+++ b/src/include/LSP/Utils.hpp
@@ -10,7 +10,6 @@
 
 std::optional<std::string> getParentPath(const std::string& path);
 std::optional<std::string> getAncestorPath(const std::string& path, const std::string& ancestorName);
-size_t pathDistance(const std::string& path1, const std::string& path2);
 std::string convertToScriptPath(const std::string& path);
 std::string codeBlock(const std::string& language, const std::string& code);
 std::optional<std::string> readFile(const std::filesystem::path& filePath);
--- a/src/operations/Completion.cpp
+++ b/src/operations/Completion.cpp
@@ -199,73 +199,169 @@ static std::vector<std::string> getServiceNames(const Luau::ScopePtr& scope)
     return services;
 }
 
+static lsp::CompletionItem createSuggestService(const std::string& service, size_t lineNumber)
+{
+    auto importText = "local " + service + " = game:GetService(\"" + service + "\")\n";
+
+    lsp::CompletionItem item;
+    item.label = service;
+    item.kind = lsp::CompletionItemKind::Class;
+    item.detail = "Auto-import";
+    item.documentation = {lsp::MarkupKind::Markdown, codeBlock("lua", importText)};
+    item.insertText = service;
+    item.sortText = SortText::AutoImports;
+
+    lsp::Position placement{lineNumber, 0};
+    item.additionalTextEdits.emplace_back(lsp::TextEdit{{placement, placement}, importText});
+
+    return item;
+}
+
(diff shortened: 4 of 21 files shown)


JohnnyMorganz (commented):
Thank you very much for tackling this! It is a longstanding feature request.

A few things:
- Can we support non-roblox code too? You can run a directory iterator to find all relevant files to use https://en.cppreference.com/w/cpp/filesystem/recursive_directory_iterator. One thing we will want to consider, however, is the performance impact
- How do we prioritise multiple modules with the same name, could we do something clever there? Say there is a `Utils` module next to the current script we are in, and also a general one in ReplicatedStorage - we should prioritise the nearby one first, before the global one
- We need to respect ignored directories: we don't want to be autocompleting modules inside of `Packages/_Index/**` for example
- Maybe the method of require should be configured (absolute vs. relative). I know Roblox LSP has some configuration here, it might be useful to follow something similar. Alternatively, check out rust-analyzer
- The line where to add the require is a bit more involved. We should place it in the correct require group (e.g., if we are auto importing `ReplicatedStorage.Packages.Roact` it should be next to other `ReplicatedStorage.Packages` requires, and `ReplicatedStorage.Shared` should be separate etc.) Maybe we can look at all require groups in our visitor, then place it next to the ones with the longest common prefix

src/operations/Completion.cpp
```diff
@@ -234,38 +270,53 @@ void WorkspaceFolder::suggestImports(
...
                 if (definedService < service && location >= lineNumber)
                     lineNumber = location + 1;
             }
 
-            if (isAlreadyDefined)
+            result.emplace_back(createSuggestService(service, lineNumber));
+        }
+    }
+
+    {
```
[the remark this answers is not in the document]
HawDevelopment: Seperate service imports from require imports

HawDevelopment:
Implementation of base lua requires using strings are out of the scope for this pull request.

src/operations/Completion.cpp
```diff
@@ -234,38 +281,63 @@ void WorkspaceFolder::suggestImports(
...
-            item.detail = "Auto-import";
-            item.documentation = {lsp::MarkupKind::Markdown, codeBlock("lua", importText)};
-            item.insertText = service;
-            item.sortText = SortText::AutoImports;
+        for (auto& [path, node] : fileResolver.virtualPathsToSourceNodes)
+        {
+            if (path == moduleName || node->className != "ModuleScript" || visitor.contains(node->name))
+                continue;
+            auto scriptFilePath = node->getScriptFilePath();
+            if (scriptFilePath.has_value() && isIgnoredFile(scriptFilePath.value(), config))
```
JohnnyMorganz, line 308: Need to use `WorkspaceFileResolver::getRealPathFromSourceNode()` here, since it handles correct relative paths
```suggestion
            if (auto scriptFilePath = fileResolver.getRealPathFromSourceNode(node); scriptFilePath && isIgnoredFile(scriptFilePath.value(), config))
```

src/Utils.cpp
```diff
@@ -44,6 +44,37 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
...
+    std::string output = "";
+    for (auto& it = p.begin(); it != p.end(); ++it)
+    {
+        auto str = it->string();
+        if (str.find(" ") != std::string::npos)
+            output += "[\"" + str + "\"]";
+        else if (str == ".")
+            output += "script";
+        else if (str == "..")
+            output += "script.Parent";
```
JohnnyMorganz, line 68: I don't think this is completely right: we should only do `script` and `script.Parent` if this is at the start of the path. If its in the middle, it should just be empty or `.Parent`. (Not sure if that case would ever be hit in practice though)

src/Utils.cpp
```diff
@@ -44,6 +44,37 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
...
 }
 
+size_t pathDistance(const std::string& path1, const std::string& path2)
+{
+    std::filesystem::path p1(path1);
+    std::filesystem::path p2(path2);
+
+    auto distance = std::filesystem::relative(p1, p2);
+    return std::distance(distance.begin(), distance.end()) - 1;
+}
```
JohnnyMorganz, line 54: I think you can set the arguments as a path directly rather than having to create another variable
```suggestion
size_t pathDistance(const std::filesystem::path& path1, const std::filesystem::path& path2)
{
    auto distance = std::filesystem::relative(path1, path2);
    return std::distance(distance.begin(), distance.end()) - 1;
}
```

src/operations/Completion.cpp
```diff
@@ -234,38 +281,63 @@ void WorkspaceFolder::suggestImports(
...
+                continue;
 
-            lsp::Position placement{lineNumber, 0};
-            item.additionalTextEdits.emplace_back(lsp::TextEdit{{placement, placement}, importText});
+            std::string requirePath;
+            auto distance = pathDistance(path, moduleName);
+            if (distance == 0)
+                requirePath = "./" + node->name;
+            else if (distance < 5)
```
JohnnyMorganz, line 315: Is there a reason about the magic number 5?

JohnnyMorganz (commented):
This is awesome, thank you!

There is a minor bug where the file name has a space in it: maybe we should replace the space with an `_` (or maybe there should be a particular casing like camel/pascal, but that's a bit too complex)


https://user-images.githubusercontent.com/19635171/224502110-d33df042-9215-4c01-9d2c-45a5f3bb3efd.mp4

In this case, could we maybe do a few things:

https://user-images.githubusercontent.com/19635171/224502211-3d981333-f0b3-4ba8-b991-8414d0b61679.mp4


1) I think requires should always be below GetService calls
2) Could it maybe build off existing variables instead, so the require uses `ReplicatedStorage`?
3) If `ReplicatedStorage` wasn't already specified, maybe we should also auto import ReplicatedStorage? I'm not too sure about this one though

Will stress test a bit more

src/operations/Completion.cpp
HawDevelopment: Nope

src/Utils.cpp
HawDevelopment: This is to be more in line with the other methods in the utils file.
```cpp
std::optional<std::string> getParentPath(const std::string& path);
std::optional<std::string> getAncestorPath(const std::string& path, const std::string& ancestorName);
size_t pathDistance(const std::string& path1, const std::string& path2);
std::string convertToScriptPath(const std::string& path);
```

src/operations/Completion.cpp
HawDevelopment: I'll make this a config setting.

HawDevelopment:
> If ReplicatedStorage wasn't already specified, maybe we should also auto import ReplicatedStorage? I'm not too sure about this one though

This is way too complicated for such simple feature. Not even Roblox LSP has this.

src/operations/Completion.cpp
```diff
@@ -199,73 +199,169 @@ static std::vector<std::string> getServiceNames(const Luau::ScopePtr& scope)
     return services;
 }
 
+static lsp::CompletionItem createSuggestService(const std::string& service, size_t lineNumber)
+{
+    auto importText = "local " + service + " = game:GetService(\"" + service + "\")\n";
```
matthargett, line 204: Use the configured line endings for the workspace instead of the hardcoded end of line here?

matthargett (commented):
Can you cover the changes in the Completion file with tests like you did for the Utils file?

src/operations/Completion.cpp
HawDevelopment: Sure, but I think this should be a separate PR. I was just doing some refactoring and moved it out of the for loop.

HawDevelopment: Adding on to this, there's hard coded line endings everywhere in the code base and is therefore not related to this PR.

HawDevelopment:
> Can you cover the changes in the Completion file with tests like you did for the Utils file?

There are no tests for Completion?

src/operations/Completion.cpp
```diff
@@ -199,73 +199,169 @@ static std::vector<std::string> getServiceNames(const Luau::ScopePtr& scope)
...
+
+    if (Luau::startsWith(path, "game/")) {
+        auto parts = Luau::split(path, '/');
+        if (parts.size() > 2) {
+            auto service = std::string(parts[1]);
+            if (visitor.serviceLineMap.find(std::string(service)) != visitor.serviceLineMap.end()) {
+                return service + "/" + Luau::join(std::vector(parts.begin() + 2, parts.end()), "/");
+            }
+        }
+    }
```
JohnnyMorganz, line 261: It would probably be useful for more than just services, for example the Roblox style guide recommends having a common ancestor defined as a variable: https://roblox.github.io/lua-style-guide/#requiring-libraries

```lua
local Packages = ReplicatedStorage.Packages
local MyPackage = require(Packages.MyPackage)
```

I think we can leave it out of this PR though if you want, and it can be done later

JohnnyMorganz:
Could you take a look at the build failures on Linux and MacOS?

src/operations/Completion.cpp
HawDevelopment: The only problem with this common ancestor approach is that it's hard to know if it's Packages or if it's a table or boolean, the best chance we have is guessing. I think the current approach being optimizing service requires if present is good enough. This feature was never meant to output the most correct require path.

HawDevelopment:
> Could you take a look at the build failures on Linux and MacOS?

Fixed!

nezuo:
I'm very excited for this! Here are some things I noticed while testing it:
- Requires aren't always grouped with the other requires. (In this example, I auto-imported Angle.)
image
- I really wish it would auto-import services. I never use `game.Service` in my requires.
- It seems to always use relative require paths if your script and the module you're requiring are in the same service. Personally, I use absolute requires unless the required module is a descendant of the script it's required in.

HawDevelopment:
> Requires aren't always grouped with the other requires. (In this example, I auto-imported Angle.)

I'll look into a better way of sorting requires.

> I really wish it would auto-import services. I never use game.Service in my requires.

This seems pretty normal. I will add a config setting for this.

> It seems to always use relative require paths if your script and the module you're requiring are in the same service. Personally, I use absolute requires unless the required module is a descendant of the script it's required in.

This is probably much better than the current 'path distance' solution. However, I think it should also be instances with the same parent.

JohnnyMorganz:
A prebuild extension is available at https://github.com/JohnnyMorganz/luau-lsp/actions/runs/4431510063 to test out.
Will look to get this merged over the weekend

nezuo:
The latest changes have some issues:
- When the service doesn't exist, it puts the module in the wrong spot:
image
- This time the service did exist and it put it before:
image
- If Model is using the ReplicatedStorage variable, it doesn't put angle in the correct order (this isn't a huge deal because StyLua can sort it on save):
image

src/include/LSP/ClientConfiguration.hpp
```diff
@@ -90,10 +90,12 @@ struct ClientCompletionConfiguration
     bool addTabstopAfterParentheses = true;
     /// If parentheses are added, fill call arguments with parameter names
     bool fillCallArguments = true;
+    /// The max 'path distance' before an auto-import require is relative
+    size_t maxRequirePathDistance = 4;
 };
 
 NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(
-    ClientCompletionConfiguration, enabled, suggestImports, addParentheses, addTabstopAfterParentheses, fillCallArguments);
+    ClientCompletionConfiguration, enabled, suggestImports, addParentheses, addTabstopAfterParentheses, fillCallArguments, maxRequirePathDistance);
```
JohnnyMorganz, line 98: I believe this setting is no longer used and should be removed

src/include/LSP/Utils.hpp
```diff
@@ -10,6 +10,8 @@
 
 std::optional<std::string> getParentPath(const std::string& path);
 std::optional<std::string> getAncestorPath(const std::string& path, const std::string& ancestorName);
+size_t pathDistance(const std::string& path1, const std::string& path2);
```
JohnnyMorganz, line 13:
```suggestion
```
Since there is no longer a definition for it

src/operations/Completion.cpp
```diff
@@ -199,73 +200,188 @@ static std::vector<std::string> getServiceNames(const Luau::ScopePtr& scope)
...
+        for (auto& [path, node] : fileResolver.virtualPathsToSourceNodes)
         {
-            // ASSUMPTION: if the service was defined, it was defined with the exact same name
-            bool isAlreadyDefined = false;
+            auto name = node->name;
+            replaceAll(name, " ", "_");
+
+            if (path == moduleName || node->className != "ModuleScript" || visitor.contains(name))
+                continue;
+            auto scriptFilePath = node->getScriptFilePath();
```
JohnnyMorganz, line 334:
```suggestion
```
Unused line

JohnnyMorganz (approved):
Going to merge it in, and any problems we can solve in future fixes.

Thank you very much for taking the time to implement this!

PR #377, opened 2023-06-06
Add SharedTable definition

### LINKED ISSUE #376: Add SharedTable to LSP
Roblox recently introduced SharedTables as a new datatype which currently doesn't exist in the current release of luau-lsp.
Roblox's built in SharedTable type type lacks autocomplete for properties, so it'd be nice if it was to be typed with generics.

JohnnyMorganz:
Closes #376 

TODO: we need to define an indexer on SharedTable


JohnnyMorganz:
This cannot be merged until the new version is released and the `LuauParseDeclareClassIndexer` FFlag is enabled

PR #411, opened 2023-07-01
Fix bulk CFrame operation types

Qualadore:
```lua
--!strict
local a = CFrame.new()
local b = Vector3.new()
print(a:PointToObjectSpace(b, b))
print(a:PointToWorldSpace(b, b))
print(a:ToObjectSpace(a, a))
print(a:ToWorldSpace(a, a))
print(a:VectorToObjectSpace(b, b))
print(a:VectorToWorldSpace(b, b))
```

### DIFF
--- a/scripts/dumpRobloxTypes.py
+++ b/scripts/dumpRobloxTypes.py
@@ -202,6 +202,14 @@ IGNORED_MEMBERS = {
         "Part0",
         "Part1",
     ],
+    "CFrame": [
+        "PointToObjectSpace",
+        "PointToWorldSpace",
+        "ToObjectSpace",
+        "ToWorldSpace",
+        "VectorToObjectSpace",
+        "VectorToWorldSpace",
+    ]
 }
 
 # Extra members to add in to classes, commonly used to add in metamethods, and add corrections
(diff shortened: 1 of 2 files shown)


JohnnyMorganz (changes requested):
Please also add them to the "IGNORED_MEMBERS" list so that the old methods are deleted

JohnnyMorganz:
I wonder if there is any way to type it so that if there are X CFrames passed as arguments, there will only be X return types. But I think this should be fine for now

Qualadore:
> I wonder if there is any way to type it so that if there are X CFrames passed as arguments, there will only be X return types. But I think this should be fine for now

I thought about generating function overloads up to the argument limit, or there could be syntax in the script for doing this. In general I think you'd need dependent types for that level of safety.

JohnnyMorganz:
Just an update on why I haven't actually merged this yet.
Merging this will cause documentation for CFrame to break, since the docs expect single-parameter CFrame methods (https://raw.githubusercontent.com/MaximumADHD/Roblox-Client-Tracker/roblox/api-docs/en-us.json).

I wonder if its possible to upstream this to Roblox somehow so that they update their docs, and hence this won't break

Qualadore:
> I wonder if its possible to upstream this to Roblox somehow so that they update their docs, and hence this won't break

You have to post the topic as a private message to the Devforum [bug-support](https://devforum.roblox.com/g/bug-support) group if you don't have Regular permissions. I've done that now, may take days or a week or more to be posted and a while to be fixed.

Qualadore:
https://devforum.roblox.com/t/bulk-cframe-operations-are-undocumented/2542278

JohnnyMorganz:
I think this was actually accidentally fixed after Roblox LSP changed their type definitions.

The documentation part is still an issue, but let's see if it gets fixed anytime soon.

PR #419, opened 2023-07-08
(feat) Plugin Rewrite - New Settings Page 

### LINKED ISSUE #313: Plugin: Support filtering what is sent to server (to prevent Payload Too Large)
Hi,

I just downloaded the studio plugin to provide DataModel info for a partially-managed Rojo project but when I attempt to connect I get this response:
image

I'm running:
Windows 11
Luau Language Server Extension v1.17.1
VSCode v1.76.0
Roblox Studio Version 0.565.0.5650363

Barocena:
Closes #313 

I added a settings page with `port`, `startAutomatically` and `include` settings

instead of plugin widget I used a module stored in analytics service (just like roblox-lsp plugin) to store settings for each place 

image


plugin/src/init.server.lua
```diff
@@ -2,24 +2,65 @@
...
+	end)
+	if
+		type(Settings) ~= "table"
+		or type(Settings.port) ~= "number"
+		or type(Settings.startAutomatically) ~= "boolean"
+		or type(Settings.include) ~= "table"
+	then
+		Settings = nil
+		warn("[Luau Language Server] Could not load settings: invalid settings")
+	end
```
JohnnyMorganz, line 52: Maybe this should provide better a more detailed message than just "invalid settings"? In particular, which setting is invalid

plugin/src/init.server.lua
```diff
@@ -2,24 +2,65 @@
...
+local function LoadSettings()
+	local result, parseError: any = loadstring(SettingsModule.Source)
+	if result == nil then
+		warn("[Luau Language Server] Could not load settings: " .. parseError)
+		Settings = nil
+		return
+	end
+	pcall(function()
+		Settings = result()
+	end)
```
JohnnyMorganz, line 43: Is there any error that can occur here which should be appropriately shown to the user?

JohnnyMorganz (commented):
My only concern with doing settings like this is that it Studio pops up with a warning about Script Injection which looks fishy. But I guess it's a reasonable compromise for now

JohnnyMorganz:
Oh also, `Settings` can be `nil` at times - how does the connect handle this? It doesn't seem to fallback to any defaults

plugin/src/init.server.lua
Barocena: I didn't encountered anything here when testing and I also handle the settings error on watchChanges before connecting

Barocena:
> Oh also, `Settings` can be `nil` at times - how does the connect handle this? It doesn't seem to fallback to any defaults

with the last commit it handles nil before connection.
what would be the best way to fallback  to default on this ? 
I didn't want it to change settings to default if there is an type error so currently, it does not connect if settings type is not correct.

plugin/src/init.server.lua
JohnnyMorganz: Is it necessary for this to be wrapped in a pcall then, if no error occurs?

plugin/src/init.server.lua
```diff
@@ -145,73 +169,24 @@ function connectServer()
...
+end
 
-widget:GetPropertyChangedSignal("Enabled"):Connect(function()
-	button:SetActive(widget.Enabled)
+SettingsModule.Changed:Connect(function()
+	if connected.Value then
+		print("[Luau Language Server] Disconnecting from DataModel changes")
+		cleanup()
+	end
+	LoadSettings()
```
JohnnyMorganz, line 189: Should we automatically reconnect if we were originally connected?

Barocena: We can check if user was connected when opened the settings module and auto reconnect, only downside is it spams to output a bit since event triggers on every character change which is why I didn't do at the first place

JohnnyMorganz: Oh I didn't realise it was every character change. Thought it was on save, but now I remember there isn't really such thing in studio

Barocena: Looks like it doesn't cause any issue other than output spamming so I think we can solve with adding a silent mode to connect function (just for auto reconnect) which is fine since we would still have warnings for invalid setting types

plugin/src/init.server.lua
Barocena: did some tests, we already handle errors on 37 but, that pcall solves a really edge case like if person put `error()` on the setting and reload the session they can't open the setting menu again, so I think we can keep the pcall

JohnnyMorganz: If we intend to keep it, we should probably re-emit the error as a warning, so that it isn't silently caught and the user knows why it's not working

JohnnyMorganz (approved):
Thank you!

PR #421, opened 2023-07-12
Fix Actor:SendMessage

Aloroid:
Incorrectly typed as `function SendMessage(self, topic: string, message: any): nil` (message should be ...any)


JohnnyMorganz:
Can you also regenerate the file?

Aloroid:
Should work now

JohnnyMorganz (approved):
Thanks

PR #423, opened 2023-07-15
add JetBrains ecosystem plugin (WIP) reference

Lordfirespeed:


JohnnyMorganz:
This is pretty neat, thank you!
Is the expectation that the Todo list is completed before this is available? Or is it usable in its current state

Lordfirespeed:
It's not yet usable, no - a bit more of the to-do list needs to be done (especially the points relating to the binaries).
I didn't realise quite how difficult implementing this plugin would be 😂 even just refactoring support is difficult.

JohnnyMorganz:
I don't think luau lsp has any refactoring operations. What kind of stuff does that entail?

Lordfirespeed:
It involves some specific implementation (on my part) in the plugin. It's not trivial to explain, but 
- see [here](https://plugins.jetbrains.com/docs/intellij/reference-contributor.html) for a simple properties language example
- see [here](https://plugins.jetbrains.com/docs/intellij/rename-refactoring.html) for details

Essentially, whilst VSCode implements a 'naive' renaming system that looks for particular strings and replaces them, JetBrains plugins use a more complicated 'smart' renaming system that involves a language lexer and parser to generate tokens and use them to build a 'PSI tree' which can then be traversed to find usages of the text that will *definitely* be a reference, rather than just a coincidental instance of the text.

TetsuzaemonIba:
I made one more attempt to create a plugin for JB IDEs (I missed the previous one, when searching for existing solutions) https://plugins.jetbrains.com/plugin/24957-luau/edit/versions/eap/772699
It's yet in the "private" beta, but I hope to release it in a month or so

JohnnyMorganz:
Closing as intellij-luau is now available

PR #425, opened 2023-07-16
Add paragraph break in generated documentation to ensure correct parsing

Anaminus:
Some Markdown parsers require an explicit paragraph break between a list block and the previous block. Sublime's LSP appears to be one of them.

No explicit break:

```markdown
**Parameters**
- `foo`  Description of foo.
- `bar`  Description of bar.
```

Explicit break:

```markdown
**Parameters**

- `foo`  Description of foo.
- `bar`  Description of bar.
```

This change adds a single newline after each section header to create the paragraph break necessary for list blocks to be interpreted correctly. Markdown parsers that do not require this paragraph break should continue parse normally.

### DIFF
--- a/src/DocumentationParser.cpp
+++ b/src/DocumentationParser.cpp
@@ -197,7 +197,7 @@ std::string printMoonwaveDocumentation(const std::vector<std::string>& comments)
 
     if (!params.empty())
     {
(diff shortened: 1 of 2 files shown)


JohnnyMorganz (approved):
Thanks!

PR #432, opened 2023-07-26
Occupant property should allow nil values

prooheckcp:
Currently, there's a small mistake on Roblox's API. It defines the Occupant property from both the Seats and VehicleSeats as "Humanoid" instead of "Humanoid?". This triggers TypeErrors when checking whether a seat is or isn't occupied.

image


JohnnyMorganz (commented):
Looks good! Could you regenerate the file as well?

```
python dumpRobloxTypes.py > globalTypes.d.luau
```

prooheckcp:
@JohnnyMorganz Currently the globalTypes.d.lua is using the file extension of .lua, do you want me to change it to .luau?

JohnnyMorganz:
Ah whoops, that's a typo. Please keep it as .lua

It is difficult to change right now as many tools rely on the URL pointing to the file with .lua ending

prooheckcp:
Should be updated now

JohnnyMorganz:
Did you generate the file in a PowerShell terminal? Looks like it has a BOM at the start.
You may need to generate it in command prompt directly

prooheckcp:
> Did you generate the file in a PowerShell terminal? Looks like it has a BOM at the start. You may need to generate it in command prompt directly

Yes, I initially used PowerShell to generate it. Just tried generating it with the terminal instead to see if that resolves it

prooheckcp:
> Can you merge in the current main? It should resolve the CI error

It's already updated with the main branch, not sure what's causing this

JohnnyMorganz (approved):
I've merged in latest main, looks like it works now.

Thanks!

PR #499, opened 2023-11-19
Add Batch DM update & Fixed print spam

Barocena:
This PR adds Batch update for sending DataModel info to server, current interval is 0.5
and removed the print when successfully send DM info to server, it was spamming prints in the output


plugin/src/init.server.lua
```diff
@@ -139,35 +143,44 @@ local function sendFullDMInfo(isSilent: boolean?)
...
 	cleanup()
+	local batchLoop = task.spawn(function()
+		while true do
+			task.wait(0.5)
+			if wasUpdated then
+				sendFullDMInfo()
+				wasUpdated = false
+			end
+		end
+	end)
```
JohnnyMorganz, line 167: Maybe instead of busy-waiting we can be a bit more clever about it and do it on demand?

Something along the lines of

```lua
local function deferSend()
  if sendTask then
    task.cancel(sendTask)
  end
  sendTask = task.delay(0.5, function()
    sendTask = nil
    ...
  end)
end
```

plugin/src/init.server.lua
```diff
@@ -134,19 +135,18 @@ local function sendFullDMInfo(isSilent: boolean?)
 
 	if not success then
 		warn(`[Luau Language Server] Connecting to server failed: {result}`)
+		cleanup()
 		connected.Value = false
```
JohnnyMorganz, line 139: Ah nice spot, we should be able to get rid of the .Value set since that's handled by cleanup
```suggestion
		cleanup()
```

plugin/src/init.server.lua
```diff
@@ -134,19 +135,18 @@ local function sendFullDMInfo(isSilent: boolean?)
...
 	elseif not result.Success then
 		warn(`[Luau Language Server] Sending full DM info failed: {result.StatusCode}: {result.Body}`)
 		connected.Value = false
```
JohnnyMorganz, line 142:
```suggestion
		cleanup()
```

JohnnyMorganz (approved):

plugin/src/init.server.lua
JohnnyMorganz: Whoops, looks like this suggestion wasn't applied. I'll do it manually

PR #527, opened 2024-01-11
Support floor division in Vector2 & Vector3

MaximumADHD:
Roblox's Luau implementation supports this.


scripts/dumpRobloxTypes.py
```diff
@@ -263,6 +263,7 @@
         "function __sub(self, other: Vector3): Vector3",
         "function __mul(self, other: Vector3 | number): Vector3",
         "function __div(self, other: Vector3 | number): Vector3",
+        "function __idiv(self, other: Vector3 | number): Vector3",
```
JohnnyMorganz, line 266:
```suggestion
```

Need to make it a suggestion so I can commit on mobile...

PR #549, opened 2024-02-07
SharedTable and Path2dControlPoint Type fixes

Barocena:
This PR changes the arg of SharedTable to optional and adds path2dcontrolpoint type to suppress error


JohnnyMorganz:
Thanks!

Issue #426, opened 2023-07-16
Improper and missing escaping when converting certain .json files to Luau


funwolf7:
Looking at the [code](https://github.com/JohnnyMorganz/luau-lsp/blob/main/src/Sourcemap.cpp#L89) that converts JSON to Luau, it has a few flaws.

First, the key of an object is not escaped, meaning that characters such as quotation marks can break out of the key and cause type errors or insert arbitrary types. For example, the following JSON file:
```json
{
	"foo\"] = Instance.new(\"Part\"),\n[\"bar": "baz"
}
```
gets typed as such:
![image](https://github.com/JohnnyMorganz/luau-lsp/assets/89697976/f6ecfd6e-dbc9-49ca-a1b7-614857c9ad64)

The fix to this would be to simply escape the key, like how every other string is escaped.

However, the usage of json.dump for escaping strings is problematic, as it escapes it for JSON files, not Luau. This is fine most of the time, as most of the JSON escape sequences are the same as Luau ones, but the Unicode escape sequence is not. In JSON, each UTF-16 surrogate pair is replaced with `\uXXXX`, while in Luau, each ccodepoint is replaced with `\u{XXXXX}`.

Since it doesn't use ensure_ascii, it will typically not escape Unicode characters. However, control characters MUST be escaped according to the JSON RFC, so it will always escape control characters. For the control characters that don't have a special representation, it will use the Unicode escape sequence. When Luau tries to analyze it for typechecking, it discovers an invalid escape sequence and gives the string a type of \*error-type\*. For example, the following JSON file:
```json
{
	"foo": "\u001F"
}
```
gets typed as such:
![image](https://github.com/JohnnyMorganz/luau-lsp/assets/89697976/9ac6c7d5-fb29-46e9-a6ff-dadbc600ce11)

This is a smaller bug and not likely to affect anyone, but is a bug nonetheless. A potential fix would be to, instead of using `json.dump`, manually escape the string to be safe for Luau code, like what `%q` in `string.format` does.

PR #553, opened 2024-02-18
Add support for TOML files

DervexDev:
## Following changes were made:
* Added [toml11](https://github.com/ToruNiina/toml11) parser
* Added new function `tomlValueToLuau` to both [Sourcemap.cpp](https://github.com/DervexHero/luau-lsp/blob/cd1f05517330c9b7ee3131d8ed62df14380021e4/src/Sourcemap.cpp#L132-L172) and [Sourcemap.hpp](https://github.com/DervexHero/luau-lsp/blob/cd1f05517330c9b7ee3131d8ed62df14380021e4/src/include/LSP/Sourcemap.hpp#L57) files
* Updated `readSource` function inside [WorkspaceFileResolver.cpp](https://github.com/DervexHero/luau-lsp/blob/cd1f05517330c9b7ee3131d8ed62df14380021e4/src/WorkspaceFileResolver.cpp#L166-L182) and `getScriptFilePath` function inside [Sourcemap.cpp](https://github.com/DervexHero/luau-lsp/blob/cd1f05517330c9b7ee3131d8ed62df14380021e4/src/Sourcemap.cpp#L28-L31)
* Added new `TEST_CASE` to [Sourcemap.test.cpp](https://github.com/DervexHero/luau-lsp/blob/cd1f05517330c9b7ee3131d8ed62df14380021e4/tests/Sourcemap.test.cpp#L22-L28)
* Updated [CMakeLists](https://github.com/DervexHero/luau-lsp/blob/cd1f05517330c9b7ee3131d8ed62df14380021e4/CMakeLists.txt#L79)

## Here is a quick demo:
https://github.com/JohnnyMorganz/luau-lsp/assets/78505208/7a275f79-31eb-4d3d-8206-2771c404e698


JohnnyMorganz (approved):
Would be nice if we also had a test case for the toml parsing. I think testing `readSource` function will be challenging in our current setup, but could we add a test case for the various `tomlValueToLuau`?

I should also really add some test cases for `jsonToLuau` to, to catch stuff like #426

DervexHero:
A bit late for the 1.28.0 release but I added two [test cases](https://github.com/JohnnyMorganz/luau-lsp/blob/a7db2203ce297a8756dff11042b178b88aa0002a/tests/Sourcemap.test.cpp#L39-L72), one to test all types and the second one to make sure that strings are escaped properly.

DervexHero:
I just saw the build test results and unfortunately the order of arrays and objects varies depending on the machine or OS, so we need to parse the Luau and TOML values to compare them directly, rather than comparing strings.

JohnnyMorganz:
Agh yeah that is annoying. Not sure what the best solution for that would be.

Maybe we could test by loading in actual files + a require and checking what the type system sees. Maybe that is a bit too much to ask right now though (I don't know how easy it is to set that up).

Will give a bit of a think, otherwise maybe we can just leave the tests out for now

DervexHero:
Actually I think that I've already found pretty good solution for that - we can use Luau VM.

There is header called `lua.h` that contains I think all the functions we need. First we have to create new Luau environment with `lua_newstate`. Then load the table string using `lua_load` and then iterate over that table with `lua_next`. And with that we will be able to compare it with the TOML/JSON table with no problems.

I haven't tested anything yet it's just based on what I've seen on the [Lua Reference Manual](https://www.lua.org/manual/5.1/manual.html#lua_load). I will give it a try tomorrow.

JohnnyMorganz:
This sounds a little bit overkill for the test case tbh, but we can see. I believe we currently have no dependence on Luau.VM, so not sure it's worth introducing one just for this.

If you think this is a worthwhile route, maybe it would be better to just look at the Parser level instead? We already have everything there. You can use Luau.AST to parse the string and extract a LuauAstExprTable to iterate through. (You may need to create a wellformed program to parse though - sticking a "return" before the table may be sufficient)

DervexHero:
> I believe we currently have no dependence on Luau.VM

Luau has VM built-in. It is located in `luau/VM` and there is that `lua.h` header

> If you think this is a worthwhile route, maybe it would be better to just look at the Parser level instead? We already have everything there. You can use Luau.AST to parse the string and extract a LuauAstExprTable to iterate through. (You may need to create a wellformed program to parse though - sticking a "return" before the table may be sufficient)

I will look into that tomorrow as well

DervexHero:
I looked into the Luau Ast source code today and I'm not really sure where do I start I mean how do I parse the source string. I didn't go in depth as I'm busy with my other projects so do you have any advise how to parse string that we have from the tests?

JohnnyMorganz:
We already have uses of "parse" in tests - see WorkspaceFileResolver as an example. Our helper function already sets up the SourceModule etc.

You may need to make the string into a well formed program, but stick a return or a local before the table should be sufficient

DervexHero:
Got the parser working, however now when I'm iterating over `AstExprTable` like this:
```cpp
for (const auto& item : table1->items)
{
    std::cout << "---" << std::endl;

    if (auto key = item.key->as<Luau::AstExprConstantString>())
    {
        std::cout << key->value.data << std::endl;
    }
}
```

`key` of the first item is always empty even tho that's the input table:
```cpp
Luau::AstStatBlock* block = parse(R"(
    local _ = {["1"] = 1, ["2"] = 2, ["3"] = 3, ["4"] = 4}
)");
```

and the output:
```
---

---
2
---
3
---
4
```

Did you encounter a similar issue by any chance?

JohnnyMorganz:
Hey sorry for the significant delay in taking a look at this. Given there was a significant rewrite of the internals, I went ahead and resolved the merge conflict (and restructured the code slightly)

I've also added the ast-based tests checking as discussed.

Do you mind checking that the requiring still works for you now?

DervexDev:
Good job! Everything related to TOML works perfectly!

However, 25 test cases fail on macOS:
- 22 in `MagicFunctions.test.cpp`
- 1 in `Rename.test.cpp`
- 2 in `WorkspaceFileResolver.test.cpp`

But it's unrelated to this PR. If needed I will provide more details.

JohnnyMorganz:
The test cases have some pre setup (make sure the test binary is running with the CWD set to the repo root).

But if the toml requires work in practice, that's good to know. Let's merge!

DervexDev:
Sorry for the misinformation but it doesn't actually work! It worked for me initially because I forgot to remove the symlink to the previous build that worked. Now only click-to-follow feature works but the actual completion does not. Additionally, it throws `TypeError: Unknown require` error.

JohnnyMorganz:
Ah dang. I'm a bit busy now but i can try and resolve it next weekend, unless you figure out the cause

DervexDev:
I can do that but in a few hours or tomorrow.

DervexDev:
Fixed in #634. It was an extremely small issue and as it turned out it also affected JSON.

PR #556, opened 2024-02-24
Fix `FindFirstAncestor` in non-DataModel projects

DervexHero:
This PR fixes behaviour of `FindFirstAncestor` when trying to `require` one of the children of root instance of non-DataModel project.

### 1. Project file
```json
{
  "name": "Root",
  "tree": {
    "$path": "src"
  }
}
```

### 2. Hierarchy
image

### 3. Before fix
image

### 4. After fix
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-- Fixed `FindFirstAncestor` in non-DataModel projects 
-
 ### Changed
 
 - Sync to upstream Luau 0.615
@@ -19,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Autocompletion of variables that hold a class type will now correctly have a kind of "variable" rather than "class"
 - Introduced a fix for orphaned `rojo` processes after VSCode has closed
+- `FindFirstAncestor` method in non-DataModel projects 
 
 ## [1.27.1] - 2024-01-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+- Fixed `FindFirstAncestor` in non-DataModel projects 
+
 ### Changed
 
-- Sync to upstream Luau 0.615
+- Sync to upstream Luau 0.612
 - Non-function properties will now no longer be shown by default when autocompleting a method call (e.g., `foo:bar`).
   To revert back to the original behaviour, enable `luau-lsp.completion.showPropertiesOnMethodCall`
 - Support Ubuntu 20.04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Autocompletion of variables that hold a class type will now correctly have a kind of "variable" rather than "class"
 - Introduced a fix for orphaned `rojo` processes after VSCode has closed
-- `FindFirstAncestor` method in non-DataModel projects 
+- `FindFirstAncestor` method now correctly finds the project root in non-DataModel projects 
 
 ## [1.27.1] - 2024-01-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Sync to upstream Luau 0.612
+- Sync to upstream Luau 0.615
 - Non-function properties will now no longer be shown by default when autocompleting a method call (e.g., `foo:bar`).
   To revert back to the original behaviour, enable `luau-lsp.completion.showPropertiesOnMethodCall`
 - Support Ubuntu 20.04
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -32,7 +32,23 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
     auto parentPathWithSlash = *parentPath + "/";
 
     // If the ancestor is the project root we need to rename it, applies only to non-DataModel projects
-    auto normalizedAncestorName = ancestorName == rootSourceNode->name ? "ProjectRoot" : ancestorName;
+    auto normalizedAncestorName = ancestorName;
+    if (rootSourceNode && !isDataModel(parentPathWithSlash) && ancestorName == rootSourceNode->name)
+    {
+        normalizedAncestorName = "ProjectRoot";
+
+        // Set normalizedAncestorName back to ancestorName if it's found in the parentPath before root
+        auto ancestors = split(parentPathWithSlash, '/');
+        for (int i = ancestors.size() - 1; i > 0; i--)
+        {
+            auto ancestor = ancestors[i];
+            if (ancestor == ancestorName)
+            {
+                normalizedAncestorName = ancestor;
+                break;
+            }
+        }
+    }
 
     auto ancestor = parentPathWithSlash.rfind(normalizedAncestorName + "/");
     if (ancestor != std::string::npos)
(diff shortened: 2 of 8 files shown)


JohnnyMorganz:
Thanks for the PR! I wonder if we can also add a test case for this. We don't really have great testing structure for sourcemap though

DervexHero:
I think we should actually add test cases for `WorkspaceFileResolver` and `Utils` as these are the files I modified in this PR. I'm going to look into this in an hour or so.

DervexHero:
Okay so first of all I had to update doctest dependency from 2.4.8 to 2.4.11 because I was getting `sprintf` deprecation error on macOS.

I fixed existing test cases and added new one for [Utils](https://github.com/JohnnyMorganz/luau-lsp/blob/f87cab44e9df5cfd19e82134ec544aa079de34fd/tests/Utils.test.cpp). I also wanted to add test case for `WorkspaceFileResolver` but I failed as I was getting "SIGSEGV - Segmentation violation signal" error at `Luau::AstStatBlock` parsing level.
Here is code I used if it is any helpful:
```cpp
TEST_CASE_FIXTURE(Fixture, "resolveModule handles FindFirstAncestor")
{
    WorkspaceFileResolver fileResolver;

    Luau::ModuleInfo baseContext{"ProjectRoot/Bar"};

    Luau::AstStatBlock* block = parse(R"(
        local t = node:FindFirstAncestor("Foo")
    )");
    REQUIRE(block != nullptr);
    REQUIRE(block->body.size > 0);

    Luau::AstStatLocal* local = block->body.data[0]->as<Luau::AstStatLocal>();
    REQUIRE(local != nullptr);
    REQUIRE_EQ(1, local->values.size);

    auto resolved = fileResolver.resolveModule(&baseContext, local->values.data[0]);

    REQUIRE(resolved.has_value());
    CHECK_EQ(resolved->name, "ProjectRoot");
}
```

src/Utils.cpp
```diff
@@ -31,15 +31,17 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
     // Append a "/" to the end of the parentPath to make searching easier
     auto parentPathWithSlash = *parentPath + "/";
 
+    // If the ancestor is the project root we need to rename it, applies only to non-DataModel projects
+    auto normalizedAncestorName = ancestorName == rootSourceNode->name ? "ProjectRoot" : ancestorName;
```
JohnnyMorganz, line 35: Root source node could be a null pointer, I think that needs to be handled appropriately.

tests/Utils.test.cpp
```diff
@@ -1,31 +1,51 @@
...
-    CHECK_EQ(getAncestorPath("game/ReplicatedStorage/Module/Child/Foo", "game"), "game");
+    auto sourceNode = std::make_shared<SourceNode>(SourceNode{});
+
+    CHECK_EQ(getAncestorPath("game/ReplicatedStorage/Module/Child/Foo", "game", sourceNode), "game");
+}
+
+TEST_CASE("getAncestorPath handles when ancestor is root of non-DataModel node")
+{
+    auto sourceNode = std::make_shared<SourceNode>(SourceNode{
+        .name = "Foo",
```
JohnnyMorganz, line 31: This requires C++20, the build is currently only on C++17

tests/Utils.test.cpp
```diff
@@ -1,31 +1,51 @@
...
-TEST_CASE("getAncestorPath handles when ancestor is root")
+TEST_CASE("getAncestorPath handles when ancestor is root of DataModel node")
 {
-    CHECK_EQ(getAncestorPath("game/ReplicatedStorage/Module/Child/Foo", "game"), "game");
+    auto sourceNode = std::make_shared<SourceNode>(SourceNode{});
+
+    CHECK_EQ(getAncestorPath("game/ReplicatedStorage/Module/Child/Foo", "game", sourceNode), "game");
+}
+
+TEST_CASE("getAncestorPath handles when ancestor is root of non-DataModel node")
```
JohnnyMorganz, line 28: You should also have a testcase when sourcenode is null

src/Utils.cpp
JohnnyMorganz, line 35: Also, I think this check is applied too early.
Consider an example scenario of:
```
Foo
- Bar
   - Foo
      - Baz <- this calls FindFirstAncestor("Foo")
```
The returned path should be `ProjectRoot/Bar/Foo`, but this will return `ProjectRoot` since `normalizedAncestorName` will equal `ProjectRoot` rather than `Foo`.

Maybe this is resolved if we just do the check after the initial attempt?

Also, we should probably first add a check to see `startsWith(path, "ProjectRoot/")` so that we don't unknowingly apply this to DataModel projects (maybe that can be extracted out to a helper function "isDatamodelProject")

tests/Utils.test.cpp
```diff
@@ -1,31 +1,51 @@
 #include "doctest.h"
 #include "LSP/Utils.hpp"
+#include "LSP/Sourcemap.hpp"
 
 TEST_SUITE_BEGIN("UtilsTest");
 
 TEST_CASE("getAncestorPath finds ancestor from given name")
 {
-    CHECK_EQ(getAncestorPath("game/ReplicatedStorage/Module/Child/Foo", "Module"), "game/ReplicatedStorage/Module");
+    auto sourceNode = std::make_shared<SourceNode>(SourceNode{});
```
JohnnyMorganz, line 9: Rather than creating sourcenodes for all these tests, can we just pass `nullptr`?

JohnnyMorganz:
>  I also wanted to add test case for WorkspaceFileResolver but I failed as I was getting "SIGSEGV - Segmentation violation signal" error at Luau::AstStatBlock parsing level.
Here is code I used if it is any helpful:

Yeah, this is because rootSourceNode is null in this test case which is currently not checked against as mentioned above. If you want to add a test case like this, you will also need to create a sourcenode. (Also, would be nice to have a sourcenode == nullptr version too)

DervexHero:
Okay did everything.

`getAncestorPath` will now handle properly `ProjectRoot/Bar/Foo/Baz`. I also added test case for this. Removed designated initializer so tests work with C++ 17 again. Added three new Util functions : `isDataModel` - with test case, `startsWith` and `split`. And finally I added working test case for `WorkspaceFileResolver`.

> You should also have a testcase when sourcenode is null

I don't think we need to add special test case for this as most of them pass `nullptr` now.

JohnnyMorganz:
> `startsWith` and `split`

We already have these functions, they are available in Luau's StringUtils (`Luau::startsWith` and `Luau::split`)

src/Utils.cpp
```diff
@@ -31,15 +31,33 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
     // Append a "/" to the end of the parentPath to make searching easier
     auto parentPathWithSlash = *parentPath + "/";
 
+    // If the ancestor is the project root we need to rename it, applies only to non-DataModel projects
+    auto normalizedAncestorName = ancestorName;
+    if (rootSourceNode && !isDataModel(parentPathWithSlash) && ancestorName == rootSourceNode->name)
+    {
+        normalizedAncestorName = "ProjectRoot";
+
+        // Set normalizedAncestorName back to ancestorName if it's found in the parentPath before root
```
JohnnyMorganz, line 40: What I meant before is that I think it is simpler to just move this check to after the initial ancestor rfind.
So it is along the lines of

```cpp
// as original
auto ancestor = parentPathWithSlash.rfind(ancestorName+ "/");
if (ancestor != std::string::npos)
{
        ...
}

// at this point we know there is definitely no ancestor with the same name within the path
// we can do a simple check and return ProjectRoot if we see that rootSourceNode.name == ancestorName

if (rootSourceNode && !isDataModel(parentPathWithSlash) && ancestorName == rootSourceNode->name)
{
    return "ProjectRoot"
}
```

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Autocompletion of variables that hold a class type will now correctly have a kind of "variable" rather than "class"
 - Introduced a fix for orphaned `rojo` processes after VSCode has closed
+- `FindFirstAncestor` method in non-DataModel projects
```
JohnnyMorganz, line 20:
```suggestion
- `FindFirstAncestor` method now correctly finds the project root in non-DataModel projects 
```

JohnnyMorganz (approved):
Thank you!

DervexHero:
> > `startsWith` and `split`
> 
> We already have these functions, they are available in Luau's StringUtils (`Luau::startsWith` and `Luau::split`)

Oh didn't noticed that it was hidden inside a namespace, sorry for that!

src/Utils.cpp
DervexHero: Oh yeah that makes way more sense

PR #558, opened 2024-02-27
Redefine RaycastResult with generics.

MaximumADHD:
Normally `RaycastResult.Instance` should be `BasePart`, but sometimes its `Attachment | Constraint` (which is why Roblox defines it as Instance).

Using generics to handle this case switch would make the typechecking less annoying.


JohnnyMorganz (approved):
Thanks!

MaximumADHD:
Yikes, I forgot `self` in GizmoRaycast. Surprised that didn’t set off any alarm.

PR #572, opened 2024-03-06
Datatype type fixes

Barocena:
This PR fixes missing/wrong datatype types
Changes:
```
- CFrame  
	- CFrame.LookAlong(at: Vector3,direction: Vector3,up: Vector3) 
	- CFrame.fromEulerAngles(rx: number,ry: number,rz: number,order: Enum.RotationOrder) 
	- CFrame.new():ToEulerAngles(order: Enum.RotationOrder):number,number,number 
- PathWayPoint 
	- PathWayPoint.new().Label (and 3rd arg of new was missing)
- PhysicalProperties 
	-  new(density: number,friction: number,elasticity: number) is missing 
- Rect 
	-  Rect.new() (params are not required) 

- Secret 
	-  Secret is not in the list at all
	-  HttpService:GetSecret():AddPrefix 
	- HttpService:GetSecret():AddSuffix 

- Vector2 
	- Vector2.max 
	- Vector2.min 
	- Vector2.new():Abs 
	- Vector2.new():Ceil 
	- Vector2.new():Floor 
	- Vector2.new():Sign 
	- Vector2.new():Angle 
- Vector3 
	- Vector3.max 
	- Vector3.min 
	- Vector3.new():Abs 
	- Vector3.new():Ceil 
	- Vector3.new():Floor 
	- Vector3.new():Sign 
```
can be tested with :
```lua
--!strict
CFrame.lookAlong(Vector3.zero,Vector3.zero)
CFrame.fromEulerAngles(1,1,1,Enum.RotationOrder.XYZ)
CFrame.new():ToEulerAngles()

local _ = PathWaypoint.new(Vector3.zero,Enum.PathWaypointAction.Jump,"custom label")

PhysicalProperties.new(1,1,1)

Rect.new()

local sc= game:GetService("HttpService"):GetSecret()
sc:AddPrefix()
sc:AddSuffix()

Vector2.max(Vector2.zero,Vector2.zero)
Vector2.min(Vector2.zero,Vector2.zero)
Vector2.new():Abs()
Vector2.new():Ceil()
Vector2.new():Floor()
Vector2.new():Sign()
Vector2.new():Angle(Vector2.zero,false)


Vector3.max(Vector3.zero,Vector3.zero)
Vector3.min(Vector3.zero,Vector3.zero)
Vector3.new():Abs()
Vector3.new():Ceil()
Vector3.new():Floor()
Vector3.new():Sign()
```
###### (vector2/vector3 max,min functions are in dump script instead of json due to some weird type generation on vararg functions)

### DIFF
--- a/scripts/dumpRobloxTypes.py
+++ b/scripts/dumpRobloxTypes.py
@@ -637,6 +637,16 @@ type HumanoidDescriptionAccessory = {
 # More hardcoded types, but go at the end of the file
 # Useful if they rely on previously defined types
 END_BASE = """
+declare Vector2: {
+	zero: Vector2,
+	one: Vector2,
+	xAxis: Vector2,
+	yAxis: Vector2,
+	new: ((x: number?, y: number?) -> Vector2),
+	max: ((...Vector2) -> Vector2),
+	min: ((...Vector2) -> Vector2),
+}
+
 declare class GlobalSettings extends GenericSettings
     Lua: LuaSettings
     Game: GameSettings
--- a/scripts/dumpRobloxTypes.py
+++ b/scripts/dumpRobloxTypes.py
@@ -637,31 +637,6 @@ type HumanoidDescriptionAccessory = {
 # More hardcoded types, but go at the end of the file
 # Useful if they rely on previously defined types
 END_BASE = """
-declare Vector2: {
-	zero: Vector2,
-	one: Vector2,
-	xAxis: Vector2,
-	yAxis: Vector2,
-	new: ((x: number?, y: number?) -> Vector2),
-	max: ((...Vector2) -> Vector2),
-	min: ((...Vector2) -> Vector2),
-}
-
-declare Vector3: {
-	zero: Vector3,
-	one: Vector3,
-	xAxis: Vector3,
-	yAxis: Vector3,
-	zAxis: Vector3,
-	fromNormalId: ((normal: EnumNormalId) -> Vector3),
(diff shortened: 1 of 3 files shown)


scripts/dumpRobloxTypes.py
```diff
@@ -637,6 +637,31 @@
 # More hardcoded types, but go at the end of the file
 # Useful if they rely on previously defined types
 END_BASE = """
+declare Vector2: {
```
JohnnyMorganz, line 640: Do we need to add an ignore somewhere given that we have defined an inlined definition instead?

Barocena: it seems to work fine without any ignores since we declare at the end, it overrides other declares of vector2/vector3
rest of the types in that declare (other than min/max) is exactly same as dump

JohnnyMorganz: i think we should probably put an explicit ignore for clarity

JohnnyMorganz (approved):
lgtm, would be nice to add the explicit ignore though

Barocena:
Instead of adding v2/v3 min max functions to dumptypes script, I decided to fix the type generation of variadic functions by adding new parameter type to datatypes file, if the parameter type is `Variadic` it will generate the type as `...{type}` instead of `...:{type}` 
###### (does not affect other existing types at all)

JohnnyMorganz (approved):
Thank you!

PR #576, opened 2024-03-19
Add roblox image previewing support

codesenseAye:
This adds the ability to preview `rbxasssetid` Roblox-style images. This caches all loaded images so an image is never looked up twice. The only downside of this addition is the necessity of an HTTPS library. I added a small header-only HTTPS library with some macros to disable this feature to allow easier compiling for other contributors (this macro is unsupported by some compilers but I am open to suggestions).

Credit to [@MoneySnyder](https://github.com/MoneySnyder) for the idea.
Here is a preview of the feature:

https://github.com/JohnnyMorganz/luau-lsp/assets/44715949/7f83d2ec-f8c1-4a1a-a4c3-80728c23dc9d

### DIFF
--- a/editors/code/package-lock.json
+++ b/editors/code/package-lock.json
@@ -30,7 +30,7 @@
         "ts-loader": "^9.2.8",
         "typescript": "^4.6.4",
         "webpack": "^5.76.0",
-        "webpack-cli": "^4.9.2"
+        "webpack-cli": "^4.10.0"
       },
       "engines": {
         "vscode": "^1.67.0"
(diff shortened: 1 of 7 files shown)


JohnnyMorganz:
Thank you for your contribution! Unfortunately, I do not think I will be able to accept this feature.

Luau LSP is predominately supporting the language features of Luau only. We have some specific Roblox support just for Roblox-specific Luau types and requires. Unfortunately, I do not think such a feature fits within this project. I've also so far been against adding any networking support to the server purely as I do not think it is necessary at this point in time.

I think this feature is best suited as an independent editor extension, or a separate language server if portability is useful.

PR #578, opened 2024-03-23
Add CFrame.lookAlong to each globalTypes entry

benbrimeyer:
## Problem
CFrame.lookAlong is a valid CFrame constructor recently added to the Roblox API, though it is not a valid constructor when using the lsp

## Solution
Add the new constructor to each of the globalTypes entry

It doesnt look like we require testing for the globalTypes entries, though I've attached screenshots of what my local extension build looks like with this change for clarity:
image
image


JohnnyMorganz (commented):
These files are autogenerated. Please fix this in the DataTypes.json file

JohnnyMorganz:
Actually sorry; this will get fixed by #572

benbrimeyer:
Sounds good

PR #634, opened 2024-06-02
Fix requiring `JSON` and `TOML` modules

DervexDev:
This quick fix solves this problem with the JSON and TOML files:
> Only click-to-follow feature works but the actual completion does not. Additionally, it throws TypeError: Unknown require error.

as discussed in #553.


JohnnyMorganz (approved):
Nice find, thanks!

I should really add some tests here to make sure the requiring of these files works

PR #641, opened 2024-06-05
Fix autocomplete importing . as service or not importing service

### LINKED ISSUE #638: Auto-import causes `.` import as service with relative requires, no service imports with absolute requires
This first occurs with 5ed092d9d273575f11b5ed212667320058cdedc2, and doesn't occur with 1.29.1 in any `requireStyle`.

For VS Code:
1. Set `luau-lsp.completion.imports.requireStyle` to `auto` or `alwaysRelative`.
2. Use this file structure:
luau-lsp-structure
3. In `parent`'s blank `init.luau`, start typing `child`, then auto-import it.

```lua
local . = game:GetService(".")

local child = require(script.child)
child
```

This is `default.project.json`.
```json
{
  "name": "luau-lsp-repro",
  "tree": {
    "$className": "DataModel",

    "ServerScriptService": {
      "Server": {
        "$path": "src/server"
      }
    }
  }
}
```

For absolute requires, no service will be imported.

Ketasaja:
Closes #638.

### DIFF
--- a/src/platform/roblox/RobloxCompletion.cpp
+++ b/src/platform/roblox/RobloxCompletion.cpp
@@ -372,7 +372,7 @@ void RobloxPlatform::handleSuggestImports(const TextDocument& textDocument, cons
                 }
             }
 
-            if (isRelative)
+            if (!isRelative)
             {
                 // Service will be the first part of the path
                 // If we haven't imported the service already, then we auto-import it


JohnnyMorganz (approved):
I remember reading this last weekend when trying to implement improvements to auto requires, and I was very confused how this was working. This makes a lot of sense now 😁, thanks for the find and the fix!

I really need to setup some tests for this 😅

PR #662, opened 2024-06-30
Add Path2DControlPoint types

Barocena:
This PR adds types for Path2DControlPoint datatype


JohnnyMorganz:
Going to merge #667 instead as it also has the data type definition. Thanks though!

Issue #661, opened 2024-06-29
Extension configuration for sourcemap name


TRHeliad:
It would be very useful if there was an option such as `luau-lsp.sourcemap.mapFile` that would allow you to change the name of the sourcemap file that luau lsp looks for.

I am setting up the structure for a new rojo managed project which involves multiple places and would like to enable intellisense for multiple places at once. The problem I have is that luau lsp only looks for a file specifically named `sourcemap.json` in the root directory of the workspace. I would like to configure a vscode workspace which changes the name of the sourcemap file that luau lsp looks for using a configuration. I would then use individual vscode windows with workspaces corresponding to places which are configured for specific sourcemap files.

TRHeliad:
Would it be better to have the property allow you to select a prefix for the file or just define the entire name? I was implementing this earlier and noticed that if you define the entire name it would require updating the file watchers.

JohnnyMorganz:
The file watcher will need to be modified regardless of what we choose, so I think defining the entire name is good.

We probably want to move the file watcher to the "setupWithConfiguration" inside of RobloxPlatform as well, since it is a roblox-specific thing, and we have access to the configured name at that point. Maybe we should also de-register the file watcher if the sourcemap setting was disabled

PR #665, opened 2024-07-01
Add sourcemap.sourcemapFile configuration

TRHeliad:
Adds an extension configuration for choosing the name of the sourcemap.

There is actually an issue I noticed with unregistering watched files. I am not experienced in working with an LSP so I'm not really sure why the message I am sending to unregister `didChangedWatchedFilesCapability` isn't actually stopping vscode from sending file changed messages for the sourcemap.


TRHeliad:
@JohnnyMorganz do you know anything about unregistering capabilities and why it might not be working? This isn't really a huge issue because if you reload the workspace after changing config it won't be watching.

JohnnyMorganz:
I'm not too familiar. If you set your logging level to verbose, do you get a response to your unregister capability request?

src/LanguageServer.cpp
```diff
@@ -504,25 +504,6 @@ void LanguageServer::onInitialized([[maybe_unused]] const lsp::InitializedParams
...
 
-    // Dynamically register file watchers
-    if (client->capabilities.workspace && client->capabilities.workspace->didChangeWatchedFiles &&
-        client->capabilities.workspace->didChangeWatchedFiles->dynamicRegistration)
-    {
-        client->sendLogMessage(lsp::MessageType::Info, "registering didChangedWatchedFiles capability");
-
-        std::vector<lsp::FileSystemWatcher> watchers{};
-        watchers.push_back(lsp::FileSystemWatcher{"**/.luaurc"});
-        watchers.push_back(lsp::FileSystemWatcher{"**/sourcemap.json"});
```
JohnnyMorganz, line 515: We should only move the registration of `sourcemap.json` to the roblox platform, and leave everything else the same. I think it should be fine to register multiple watcher capabilities with different IDs.

TRHeliad: Are RobloxLanguageServer and LanguageServer somehow separate with regards to capabilities? Sorry I don't really know what you mean by IDs. However, I did push changes separates the watchers so that only sourcemap.json is in the roblox platform.

TRHeliad: I checked the response it was getting an error. Looks like it was from a typo.

TRHeliad:
Looks like the LSP spec has a little typo so my parameter name was wrong. The unregistration gets a success response now, however vscode still seems to be watching the sourcemap if sourcemaps were enabled by the configuration earlier in the same session. Seems like it is out of my hands, but shouldn't be a big deal since you can reload to window if you don't want it watched.

TRHeliad:
Closes #661

JohnnyMorganz (approved):
Thanks, and so sorry for the delay!

Overall looks good to me, I did a slight change to make the ID different that we register for sourcemap watching compared to standard watching. Since we don't want to deregister the default watcher when we actually just want to deregister sourcemap watching.

Do you mind giving a quick check to see if this still works? Otherwise I can merge it in

TRHeliad:
Just pushed a fix I forgot to add. Tested and working for me.

JohnnyMorganz (approved):
Thanks!

PR #667, opened 2024-07-03
Fully Add Path2DControlPoint

MaximumADHD:
This adds Path2DControlPoint with its constructor and type definition.


JohnnyMorganz (approved):
Thanks!

PR #678, opened 2024-07-11
Add Instance.fromExisting type

Barocena:
this PR adds `Instance.fromExisting` type


JohnnyMorganz (approved):
Thanks!

We might want to attach a magic function to this to provide better types

JohnnyMorganz:
Magic function added in 20caa866cbc0a2fbd6570ab9704d81a7d76aa3bd

PR #714, opened 2024-07-24
Add plugin feedback for plugin-server connection

ExtremeKiller145:
I think it was mentioned like a week ago that plugin connection feedback would probably be added eventually, so I propose a quick fix here.

Only changes made are the added message, and passing the isSilent parameter to use it as a condition for the message.

Keep in mind I haven't actually tested this, I just read the plugin code and coded the message where I think is most appropriate, since setting up the plugin testing environment is took much of a headache for me, since I've never done it.

At the very least if my implementation sucks, this is a reminder that this 'feature' should be simple to implement, and to put it in the to-do list.

### DIFF
--- a/plugin/src/init.server.lua
+++ b/plugin/src/init.server.lua
@@ -119,7 +119,7 @@ local function cleanup()
 	connected.Value = false
 end
 
-local function sendFullDMInfo()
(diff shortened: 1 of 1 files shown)


JohnnyMorganz (approved):
Nice addition, thanks!

PR #744, opened 2024-08-16
Fix platform being initialized twice

voidedWarranties:
Introduced in https://github.com/JohnnyMorganz/luau-lsp/commit/e8b31d3d8b8b06d3a25891acaa722ed55c883f10


JohnnyMorganz (approved):
Thanks!

PR #745, opened 2024-08-16
Fix loading type definitions that rely on mutation

### LINKED ISSUE #658: Type definition file stopped working after 1.30.0
I updated the extension and my definitions file fails to read
it works again when i revert to 1.29.1
settings.json:
```json
{
    "luau-lsp.sourcemap.enabled": false,
    "luau-lsp.platform.type": "roblox",
    "luau-lsp.types.definitionFiles": [
        "./build/pilot.d.lua"
    ],
}
```
definitions file:
[pilot.d.lua](https://github.com/user-attachments/files/15992633/pilot.d.txt)

voidedWarranties:
Fixes #658 

Changes:

- Previously, the workspace-global definitions file metadata (`definitionsMetadata`) was set to the metadata of the last definitions file that contained metadata. Now, it is set to the first one. I think this makes sense as it prevents the platform metadata from getting overridden.
- Since the language server does not do anything useful before the platform is available, `WorkspaceFolder::initialize` (which only performed type registration) was moved to `setupWithConfiguration` and renamed to `registerTypes`.
- The platform mutation function is now run after every definition file is registered. The platform can disambiguate which definitions file is which by checking what types are available (how it was done before) or with the metadata provided in the file.

This should more or less match the behavior from before 1.30.0. I tested with pilot.d.lua and it seems to work.

### DIFF
--- a/src/Workspace.cpp
+++ b/src/Workspace.cpp
@@ -278,7 +278,7 @@ void WorkspaceFolder::indexFiles(const ClientConfiguration& config)
     client->sendTrace("workspace: indexing all files COMPLETED");
 }
 
-void WorkspaceFolder::initialize()
+void WorkspaceFolder::registerTypes()
 {
     client->sendTrace("workspace initialization: registering Luau globals");
     Luau::registerBuiltinGlobals(frontend, frontend.globals, /* typeCheckForAutocomplete = */ false);
@@ -304,7 +304,8 @@ void WorkspaceFolder::initialize()
 
         // Parse definitions file metadata
         client->sendTrace("workspace initialization: parsing definitions file metadata");
-        if (auto metadata = types::parseDefinitionsFileMetadata(*definitionsContents))
+        auto metadata = types::parseDefinitionsFileMetadata(*definitionsContents);
+        if (!definitionsFileMetadata)
             definitionsFileMetadata = metadata;
         client->sendTrace("workspace initialization: parsing definitions file metadata COMPLETED", json(definitionsFileMetadata).dump());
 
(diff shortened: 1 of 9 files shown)


src/Workspace.cpp
```diff
@@ -357,10 +362,7 @@ void WorkspaceFolder::setupWithConfiguration(const ClientConfiguration& configur
         platform = LSPPlatform::getPlatform(configuration, &fileResolver, this);
         fileResolver.platform = platform.get();
 
-        client->sendTrace("workspace: applying platform mutations on definitions");
-
-        platform->mutateRegisteredDefinitions(frontend.globals, definitionsFileMetadata);
-        platform->mutateRegisteredDefinitions(frontend.globalsForAutocomplete, definitionsFileMetadata);
+        registerTypes();
```
JohnnyMorganz, line 365: Since this is in setupWithConfiguration, this is now going to re-register all the builtins whenever config changes, when it did not use to do that.

That is potentially a problem, can we make it only register once on startup like before?

JohnnyMorganz (commented):
Thanks for taking the time to resolve this! And sorry for the delay. Just have one comment

src/Workspace.cpp
voidedWarranties: I think this is not an issue, see context: https://github.com/JohnnyMorganz/luau-lsp/blob/a1f9302db247bd6949c98c5d8fbc6e37bcd0ef22/src/Workspace.cpp#L356-L366

JohnnyMorganz: Whoops, my bad! GitHub decided to hide those extra lines from me 🤦‍♂️

JohnnyMorganz (approved):
Gonna test this, but looks good to me!

JohnnyMorganz (approved):
I'm trying to do better with testing, so decided to add a couple of test cases here.

Thanks for taking the time to fix!

PR #751, opened 2024-08-22
Create cmake-multi-platform.yml

FugLord77:


FugLord77:
Ignore all that my bad.

PR #682, opened 2024-07-12
Support Windows ARM64

dig1t:
Adds `win32-arm64` as a platform target when publishing to vs marketplace.
Closes #673.

PR #753, opened 2024-08-24
Support-winarm64

howmanysmall:
This should just add windows arm64 support.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,10 +57,7 @@ jobs:
         include:
           - os: windows-latest
             artifact-name: luau-lsp-win64
-            code-target: [win32-x64]
-          - os: windows-latest
-            artifact-name: luau-lsp-win-arm64
-            code-target: [win32-arm64]
+            code-target: [win32-x64, win32-arm64]
           - os: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


howmanysmall (commented):
Ignore the horrendous commits, I was just testing to see if I could build it locally first.

howmanysmall:
I also don't know if I got it building for arm64, if you could check that, that would be sweet.

JohnnyMorganz:
Apologies for the delay!

This change isn't really any different than https://github.com/JohnnyMorganz/luau-lsp/pull/753/files

It is still running on a `windows-latest` runner, which is x86_64 not arm, hence it will not build for ARM

howmanysmall:
Unfortunate. Any way to get the extension building for ARM64 at all, since that would be fine even if it's emulated.

JohnnyMorganz:
Do windows ARM pcs support emulation of x86_64 binaries by default? I'm unfamiliar with this space

If so, then happy to add arm as an extra code target for the x86_64 version

howmanysmall:
yeah it's prism emulation

JohnnyMorganz:
Closed in favour of #682 as it reuses the single artifact whilst this one would require a new rebuild

PR #777, opened 2024-09-20
Use the creator docs to generate type information

### LINKED ISSUE #760: BrickColor.Random() doesn't exist in roblox types
BrickColor.Random() is an addition to the roblox BrickColor library diffrent from `random()`, as it uses roblox's random class internally for better randomness. That currently isn't defined in the type for the BrickColor library

### LINKED ISSUE #768: FuzzyEq method is missing from the CFrame data type
RobloxStudioBeta_H7GPtyDw3N

This method exists and works in Roblox Studio, and it's [documented](https://create.roblox.com/docs/reference/engine/datatypes/CFrame#FuzzyEq) as well.

JohnnyMorganz:
Uses the creator-docs (https://github.com/Roblox/creator-docs) as the source of truth for type information generation

Closes #768 
Closes #760


Ketasaja:
> It helps in some cases (particularly data types), but makes types slightly worse in quite a few other cases.

I'm fine with contributing to Creator Docs, but I don't want worse types. If there are cases where Creator Docs doesn't support better types, like optionals, this would be more annoying. Changing some class fields is cumbersome [too](https://github.com/Roblox/creator-docs/pull/842).

JohnnyMorganz:
The creator docs don't support type submissions, so this is unfortunately not worth it for now.

Maybe we can just read the DataModel type information from it.

Issue #83, opened 2022-07-22
Types a and Player cannot be compared because they do not have the same metatable


Qualadore:
No warnings occur in Studio.
```lua
--!strict
local event = game:GetService("ReplicatedStorage"):FindFirstChild("RemoteEvent")
assert(event:IsA("RemoteEvent"), "") -- LSP does now recognize `event` as a `RemoteEvent`.

-- Won't warn if `player` is typed explicitly, but `Player` type is inferred anyway.
event.OnServerEvent:Connect(function(player, target: Player)
	print(player == target) -- Warning.
end)

local working_event: RemoteEvent = game:GetService("ReplicatedStorage"):FindFirstChild("RemoteEvent")
working_event.OnServerEvent:Connect(function(player, target: Player)
	print(player == target) -- No warning.
end)
```
```
TypeError: Types a and Player cannot be compared with == because they do not have the same metatable
```

JohnnyMorganz:
This is because the typechecker actually resolves `event` as `any`, although the hover/autocomplete resolves it to a stricter type.
So typechecker doesn't actually recognise event as a RemoteEvent, so cannot infer `player` is Player.

Its a bit strange that the generic type cannot be compared. We could make the warning disappear by removing the metatable from Player, but it currently causes issues as we it no longer type checks against comparing with other types of no metatable (string/number etc.)

Qualadore:
That's interesting, I thought `event` would've been inferred as `Instance` based on #76 and hovering over it. When you mentioned the possibility of `Instance:FindFirstChild()` returning `Instance?`, would that possibly let `event` be inferred as `Instance?` instead of `any`?

If not, I'd personally much rather write `local event: Instance?` than a metatable change that makes everything less safe.

JohnnyMorganz:
The type diagnostics system and the hovering/autocomplete systems are two separate Luau typecheckers. The latter has stricter types than the former.

Unfortunately, the type diagnostics system infers all DataModel indexes as `any`: game, workspace, game.ReplicatedStorage.Modules, or FFC/WFC variants will all be typed as any in the eyes of the type diagnostics system.

[...]

PR #783, opened 2024-09-27
Refactor for new solver

checkraisefold:
**Todo:**
- [x] Fixed autocompletion
- [x] Fixed type registration and diagnostics
- [x] Fixed inlay hints
- [x] Fixed hover types/documentation
- [x] Fixed sourcemap

**Luau issues worked around:**
- [x] https://github.com/luau-lang/luau/issues/1441
- [x] https://github.com/luau-lang/luau/issues/1447

**Blockers:**
EDIT: Worked around with magic function
~~https://github.com/luau-lang/luau/issues/1447 breaks the sourcemap WaitForChild overload registration and causes those tests to fail. I don't see a conceivable workaround for this for now because it causes a type checker error and there is no alternative to the magic function here~~

**Original description:**
The new solver has unified autocompletion and diagnostic types. In addition, the nonstrict mode has been significantly improved, meaning #83 has been completely sidestepped. As such, strict datamodel type flags should be ignored with the new solver enabled, and type registration should not occur on globalsForAutocomplete (because it is ignored by Luau).

In addition, loadDefinitionFile and by extension registerBuiltinGlobals, though still taking a typeCheckForAutocomplete argument, that argument no longer does anything within Luau internals, meaning the default to false is safe and it can be removed from explicit use.

Once the new solver is fully stable, the feature flag gates can be removed and this behavior can be the default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Sync to upstream Luau master
+- Sync to upstream Luau 0.650
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Sync to upstream Luau 0.651
+- Sync to upstream Luau 0.650
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+
 ### Changed
 
-- Sync to upstream Luau 0.650
+- Sync to upstream Luau master
+
+### Fixed
+
+- Fixed inlay hints, autocompletion, type registration, hover types/documentation, and some crashes for cases where the new solver is enabled
 
 ## [1.34.0] - 2024-10-27
 
@@ -117,7 +122,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Sync to upstream Luau 0.632
-- Language clients are recommended to send configuration during intializationOptions (see https://github.com/JohnnyMorganz/luau-lsp/blob/main/editors/README.md for details)
+- Language clients are recommended to send configuration during intializationOptions (see <https://github.com/JohnnyMorganz/luau-lsp/blob/main/editors/README.md> for details)
 - Removed need for postponing requests whilst waiting for platform configuration (relies on clients sending config in intializationOptions)
 
 ### Fixed
@@ -150,7 +155,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Sync to upstream Luau 0.631
-- An indexed expression will no longer show an inlay hint if the index matches the parameter name (i.e., `call(other.value)` won't add `value: ` inlay hint) ([#618](https://github.com/JohnnyMorganz/luau-lsp/issues/618))
+- An indexed expression will no longer show an inlay hint if the index matches the parameter name (i.e., `call(other.value)` won't add `value:` inlay hint) ([#618](https://github.com/JohnnyMorganz/luau-lsp/issues/618))
 - Studio Plugin will now perform Gzip compression on sent requests
 
 ### Fixed
@@ -182,7 +187,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Sync to upstream Luau 0.625
 - Improved memory usage of document and workspace diagnostics by no longer storing type graphs
-- Rewritten the Luau grammar syntax: https://github.com/JohnnyMorganz/Luau.tmLanguage
+- Rewritten the Luau grammar syntax: <https://github.com/JohnnyMorganz/Luau.tmLanguage>
 
 ### Fixed
 
@@ -604,7 +609,7 @@ local DATA = {
 
 --- Doc comment
 type Contents = {
-	...
+ ...
 }
 ```
 
@@ -763,7 +768,7 @@ local y = tbl.data -- Should give "This is some special information"
 - Provide autocomplete for class names in `Instance:IsA("ClassName")` and errors when ClassName is unknown
 - Provide autocomplete for properties in `Instance:GetPropertyChangedSignal("Property")` and errors when Property is unknown
 - Provide autocomplete for enums in `EnumItem:IsA("enum")` and errors when Enum is unknown
-- Added support for moonwave-style documentation comments! Currently only supports comments attached to functions directly. See https://eryn.io/moonwave for how to write doc comments
+- Added support for moonwave-style documentation comments! Currently only supports comments attached to functions directly. See <https://eryn.io/moonwave> for how to write doc comments
 - Added command line flag `--ignore=GLOB` to `luau-lsp analyze` allowing you to provide glob patterns to ignore diagnostics, similar to `luau-lsp.ignoreGlobs`. Repeat the flag multiple times for multiple patterns
 
 ### Changed
@@ -981,7 +986,7 @@ local y = tbl.data -- Should give "This is some special information"
 - Fixed regression where diagnostics are not cleared when you close an ignored file
 - Fixed errors sometimes occuring when you index `script`/`workspace`/`game` for children
 - Fixed internal error caused by `:Clone()` calls when called on an expression which isn't an Lvalue (e.g., `inst:FindFirstChild(name):Clone()`)
-- Fixed bug where `_: ` would not be removed as the name of function arguments. `function foo(_: number, _: number)` will now show as `function foo(number, number)`
+- Fixed bug where `_:` would not be removed as the name of function arguments. `function foo(_: number, _: number)` will now show as `function foo(number, number)`
 - Fixed analyze mode not exiting with a non-zero exit code when there are errors
 - Fixed excessive whitespace in document symbols for expr-named function defintions
 - Fixed hover for global functions and local variables
@@ -1147,7 +1152,7 @@ local y = tbl.data -- Should give "This is some special information"
 
 - Hover over definitions will now try to give more expressive types
 - `self` will now no longer show up in hover/signature help if it has already been implicitly provided
-- `_: ` will no longer show up in hover/signature help for unnamed function parameters
+- `_:` will no longer show up in hover/signature help for unnamed function parameters
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-
 ### Changed
 
-- Sync to upstream Luau 0.650
+- Sync to upstream Luau 0.651
 
 ### Fixed
 
-- Fixed inlay hints, autocompletion, type registration, hover types/documentation, and some crashes for cases where the new solver is enabled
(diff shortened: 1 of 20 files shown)


src/LuauExt.cpp
```diff
@@ -51,11 +51,10 @@ std::optional<nlohmann::json> parseDefinitionsFileMetadata(const std::string& de
     return std::nullopt;
 }
 
-Luau::LoadDefinitionFileResult registerDefinitions(
```
JohnnyMorganz, line 54: We still need the `typeCheckForAutocomplete` here for the old solver right?

src/Workspace.cpp
```diff
@@ -313,8 +315,9 @@ void WorkspaceFolder::initialize()
         client->sendTrace("workspace initialization: parsing definitions file metadata COMPLETED", json(definitionsFileMetadata).dump());
 
         client->sendTrace("workspace initialization: registering types definition");
-        auto result = types::registerDefinitions(frontend, frontend.globals, *definitionsContents, /* typeCheckForAutocomplete = */ false);
-        types::registerDefinitions(frontend, frontend.globalsForAutocomplete, *definitionsContents, /* typeCheckForAutocomplete = */ true);
+        auto result = types::registerDefinitions(frontend, frontend.globals, *definitionsContents);
+        if (!FFlag::LuauSolverV2)
+            types::registerDefinitions(frontend, frontend.globalsForAutocomplete, *definitionsContents);
```
JohnnyMorganz, line 320: Similarly as above, this still needs `typeCheckForAutocomplete`

JohnnyMorganz (commented):
Thanks for the PR! Looks mostly good to me

> In addition, the nonstrict mode has been significantly improved, meaning https://github.com/JohnnyMorganz/luau-lsp/issues/83 has been completely sidestepped

I don't think this is completely true. We are now going to trigger false positives again in the new type checker due to mismatched types. What we need to do is implement appropriate `read/write` types to the generated DataModel types. i.e., the `read` type should be the more expressive DM type, whilst the `write` type should be the generic Roblox type, allowing you to correctly read `inst.Parent` but write anything to it.

JohnnyMorganz (commented):
Also can you add some tests to `Autocomplete.test.cpp` for this?

You might be able to just re-use the existing tests, maybe switch them to use `LUAU_BOTH_SOLVERS_TEST_CASE_FIXTURE` so that they test with both solvers

src/LuauExt.cpp
checkraisefold: See PR description:
```
In addition, loadDefinitionFile and by extension registerBuiltinGlobals, though still taking a typeCheckForAutocomplete argument, that argument no longer does anything within Luau internals, meaning the default to false is safe and it can be removed from explicit use.
```
This includes the old solver.

src/Workspace.cpp
checkraisefold: Other comment

src/LuauExt.cpp
JohnnyMorganz: Ah fair enough, didn't check Luau internals but indeed you are right

checkraisefold:
> Thanks for the PR! Looks mostly good to me
> 
> > In addition, the nonstrict mode has been significantly improved, meaning #83 has been completely sidestepped
> 
> I don't think this is completely true. We are now going to trigger false positives again in the new type checker due to mismatched types. What we need to do is implement appropriate `read/write` types to the generated DataModel types. i.e., the `read` type should be the more expressive DM type, whilst the `write` type should be the generic Roblox type, allowing you to correctly read `inst.Parent` but write anything to it.

I checked and it flat out does not work either way with both examples in #83, but in a different way. There's no warning at all, but the first argument of the connection is `*error-type*`

tests/main.cpp
```diff
@@ -1,2 +1,30 @@
...
+int main(int argc, const char** argv)
+{
+    for (int i = 1; i < argc; ++i)
+    {
+        if (argv[i][0] != '-')
+            continue;
+        if (strcmp(argv[i], "--new-solver") == 0)
+        {
+            FFlag::LuauSolverV2.value = true;
+            DFInt::LuauTypeSolverRelease.value = 645;
```
JohnnyMorganz, line 18: can we use max int here so that we don't have to keep bumping it and it will auto work for all future releases?

src/DocumentationParser.cpp
```diff
@@ -459,7 +459,7 @@ std::optional<std::string> WorkspaceFolder::getDocumentationForType(const Luau::
     {
         return printMoonwaveDocumentation(getComments(ftv->definition->definitionModuleName.value(), ftv->definition->definitionLocation));
     }
-    else if (auto ttv = Luau::get<Luau::TableType>(followedTy); ttv && !ttv->definitionModuleName.empty())
```
JohnnyMorganz, line 462: Why did we remove this? won't this break the next line where we reference `ttv->definitionModuleName`?

JohnnyMorganz:
> Also can you add some tests to Autocomplete.test.cpp for this?

Did you end up adding the DataModel types for the new solver? I don't see a test case for it

src/DocumentationParser.cpp
checkraisefold: looks accidental

checkraisefold:
All tests passing woohoo!!!!
🥳 🎉

src/operations/Completion.cpp
```diff
@@ -369,10 +371,21 @@ std::optional<std::string> WorkspaceFolder::getDocumentationForAutocompleteEntry
...
+                    auto followedTy = Luau::follow(*parentTy);
+                    if (auto propInformation = lookupProp(followedTy, name))
                         definitionModuleName = Luau::getDefinitionModuleName(propInformation->first);
                     else
-                        definitionModuleName = Luau::getDefinitionModuleName(*parentTy);
+                        definitionModuleName = Luau::getDefinitionModuleName(followedTy);
+
+                    // TODO: Dirty hack for invalid definitionModuleName on type aliases for solver v2!
+                    // Remove after https://github.com/luau-lang/luau/issues/1441 is closed!
+                    if (FFlag::LuauSolverV2 && !definitionModuleName)
```
JohnnyMorganz, line 382: Why not use the `lookupTypeDefinitionModule` function you created here as well?

src/LuauExt.cpp
```diff
@@ -428,6 +429,39 @@ lsp::Diagnostic createParseErrorDiagnostic(const Luau::ParseError& error, const
...
+    if (!FFlag::LuauSolverV2)
+    {
+        return Luau::getDefinitionModuleName(type);
+    }
+
+    type = Luau::follow(type);
+    if (auto ttv = Luau::get<Luau::TableType>(type); ttv && ttv->definitionModuleName.empty())
+    {
+        if (type->owningArena && type->owningArena->owningModule)
+            return type->owningArena->owningModule->name;
```
JohnnyMorganz, line 445: This seems incredibly hacky. Are we sure owningModule is the correct module in all cases? If a type got copied out of a module, the owning module would change

src/operations/InlayHints.cpp
```diff
@@ -157,74 +224,32 @@ struct InlayHintVisitor : public Luau::AstVisitor
...
-                        if (!param->annotation && param->name != "_")
-                        {
-                            lsp::InlayHint hint;
-                            hint.kind = lsp::InlayHintKind::Type;
-                            hint.label = ": " + Luau::toString(argType, stringOptions);
-                            hint.position = textDocument->convertPosition(param->location.end);
-                            makeInsertable(config, hint, argType);
-                            hints.emplace_back(hint);
-                        }
+        // TODO: Remove me later if this is in fact not an intended feature in the new solver
```
JohnnyMorganz, line 244: Can you explain what is happening here? I don't get why this changed

checkraisefold: Essentially, in the new solver, the astType for the AstStatFunction is incomplete with both arguments typed as `unknown` if that function is a member function of an object (e.g. `table:test()`). However, the complete type is accessible using the AstExprIndexName ast type.

See new solver AssignPropConstraint and my OSS discord discussion
image

src/LuauExt.cpp
checkraisefold: iirc my brief testing had it correct for even imported types from other modules. the problem is I couldn't find any other way to get this information given the luau bug making definitionModuleName empty.

src/operations/InlayHints.cpp
checkraisefold: Also the git diff is HORRIBLE here I recommend reading the actual file to reduce confusion if you aren't

checkraisefold:
#806 is a blocker for new solver functionality until it's patched

src/LuauExt.cpp
checkraisefold: https://github.com/luau-lang/luau/pull/1495
new solution : fix the luau bug

src/operations/InlayHints.cpp
JohnnyMorganz: I think we should also take a deeper look at Luau here before fixing. This either seems like a Luau regression, or there must be a new struct to search this data up instead of astTypes.

I'll take a look if I'm free if you don't get round to it

checkraisefold: The other alternative is this is stored in a refinement, which I think is the correct place to look instead? See scope->rvalueRefinements

checkraisefold:
Do not merge - waiting on https://github.com/luau-lang/luau/pull/1495 to remove the hacky mess workaround that is in here
Additionally waiting on https://github.com/luau-lang/luau/pull/1493 to make the roblox def files actually, you know, work instead of being errortype spam, though this isn't a blocker for this PR

CHANGELOG.md
```diff
@@ -6,9 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+
 ### Changed
 
-- Sync to upstream Luau 0.650
+- Sync to upstream Luau master
```
JohnnyMorganz, line 12: Please just put the version of the latest release instead of master

src/operations/Hover.cpp
```diff
@@ -162,6 +162,7 @@ std::optional<lsp::Hover> WorkspaceFolder::hover(const lsp::HoverParams& params)
         auto typeFun = scope->lookupType(typeName);
         if (!typeFun)
             return std::nullopt;
+
```
JohnnyMorganz, line 165: nit: redundant

src/Flags.cpp
```diff
@@ -50,6 +54,13 @@ void registerFastFlags(std::unordered_map<std::string, std::string>& fastFlags,
     {
         onWarning(std::string("Unknown FFlag: ") + key);
     }
+
+    // These are required flags for internal functionality.
```
JohnnyMorganz, line 58: We shouldn't force these enabled, since the point of FFlags here are to dynamically opt into the configuration (and have the ability to opt out if there are problems)

src/operations/InlayHints.cpp
JohnnyMorganz: Can you move this change out of this PR and into a separate one? I think we need to check what's up with Luau here first before changing it, but I don't want to block the rest of the fixes

src/Flags.cpp
checkraisefold: The documentation system actually can crash with these off and it fails like 7 tests. I would have to enable the fflags in each failing fixture as the alternative.

src/operations/InlayHints.cpp
checkraisefold: Split off into #815

checkraisefold:
The tests will fail because you had me split off the InlayHints fix and I feel like flat out turning off the tests is a bad idea

src/Flags.cpp
checkraisefold: Kept the flags forced in the testing suite as the alternative

JohnnyMorganz:
@checkraisefold Types smoketest is segfaulting, is this because of the flags? Ideally we don't crash even if the flags are not enabled

JohnnyMorganz:
It actually seems related to the recent luau update to 0.651. Going to undo that sync for now

JohnnyMorganz (approved):
Thank you!

PR #815, opened 2024-11-07
Fix inlay hints for new solver

checkraisefold:
Split from #783 

Requires further investigation - specifically in regards to AstExprIndexName hack and whether it should be swapped to use the scope->rvalueRefinements result instead


checkraisefold:
Do not merge!

JohnnyMorganz:
Turns out this is indeed a regression in the new solver, not a luau-lsp bug: https://github.com/luau-lang/luau/issues/1516

PR #817, opened 2024-11-08
Add Content Datatype types

Barocena:
this PR adds types for Content Datatype


JohnnyMorganz (approved):
Thanks!

We should probably move away from this JSON file sometime soon since there's no reason for the indirection anymore, maybe we can directly writes these in the Python script

PR #819, opened 2024-11-08
Fix isIgnoredFile throwing error if the drive letter is lowercase

### LINKED ISSUE #752: isIgnoredFiles with default constructed path due to drive letter
I'm not sure this happens cause of the sourcemap, but I did some tests and I only get this error in scripts included in the sourcemap, maybe because I'm using argon.
image

srthMD:
I got annoyed at having to downgrade to use the lsp with nvim in windows so i decided to yolo it and try to make my own fix

i dont have much experience with c++ nor am i familiar with how the lsp works so any feedback would be helpful

fixes #752


JohnnyMorganz:
Thanks for the PR! And sorry for the delay in responding here. This is pretty much what I was planning, except we probably want to switch to a lowercase instead of uppercase by default.

I've ended up fixing this in 07a953d00ce6d433926c5005d30e7bc186d6dc80

PR #831, opened 2024-11-10
Add setting to enable new solver

checkraisefold:
Many people do not want to turn on and adjust several FFlags and be up to date with the latest FFlags (primarily through having to be in a Discord server..) just to make the new solver work correctly. This mitigates the issue by just adding one setting for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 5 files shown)


JohnnyMorganz (commented):
- I think this setting should be purely on the VSCode side. We don't need to make any changes to the server here

- We should do the settings before applying FFlag overrides on the client, to allow users to override the config if necessary

JohnnyMorganz:
Also please update changelog

JohnnyMorganz (approved):
Thanks!

PR #834, opened 2024-11-13
Plugin settings now save.

TaylorsRus:
Small factor rewrite, no major logic changes but some reformatted. Plugin settings now save, namely resulting in StartAutomatically now working as intended.


TaylorsRus:
First pull request so if there's a big bad somewhere in the repo setup just let me know

PR #837, opened 2024-11-13
Plugin settings now save, minor code refactor

TaylorsRus:
Made a new pull request because I messed up the last one.
Removed default settings module in favor of string constant in one script
Gave the code a bit of a makeover and removed some redundancy
Replaced destination of module from AnalyticsService to TestService, free free to change to whatever service works but this is one of the best options we found which also works.


TaylorsRus:
Doing some QOL changes

PR #857, opened 2024-12-07
Move Content.none to the constructor table

ddavness:
Expands on #817 - `Content.none` is part of the constructor table for this DataType (and not of the type itself)

Example:
```luau
-- valid
imageLabel.ImageContent = Content.none

-- "none is not a valid member of Content"
local c = Content.fromUri("")
print(c.none == Content.none)
```


JohnnyMorganz (approved):
Thanks!

PR #860, opened 2024-12-09
Added listener for name changes, instance parent setting to nil and a…

ItzMrRatsP:
…dded default settings

First, Added listener for when name changes it will run deferSend() function which will update in **VS Code**, This will also work when parent changed or the instance is removed.

Second, Added default settings like port and include list, And added silent value.


JohnnyMorganz:
Thanks for the taking the time to create this PR!

I've implemented some different fixes in 492f7a43d3db47c509522308fae842105388dd11, which should resolve these issues

PR #861, opened 2024-12-09
Added silent value to settings.json

ItzMrRatsP:
silent value allows you to enable and disable notifications which gives you easier time debugging, There should be a button that lets you disable it.


JohnnyMorganz:
Since #860 is not being merged, and this seems related, I am going to close this PR for now.

I've added a separate setting `debug` that can be enabled in the settings file. This will enable any `debugPrint` calls in the Luau code

PR #862, opened 2024-12-11
Plugin Script Rewrite

TaylorsRus:
**Bug fixes**
- Settings now save
- Instance changes are now properly listened to and managed

**Implemented features**
- Slightly more verbose warnings 

**Planned features (future PRs, other collaborators)**
- Only send delta DataModel information, not entire DataModel for changes
- Instead of ignoring entire instances because of one or two problematic properties, just ignore the properties
- 'Debug Mode' for enabling a bunch of extra prints and warns
- Sending DM info via URL and not Port for VS Web


plugin/src/init.server.lua
```diff
@@ -1,222 +1,278 @@
...
+		warn(`[Luau Language Server] Sending DataModel info failed: {Result.StatusCode}: {Result.Body}`)
+		CleanUpConnections()
 	else
-		connected.Value = true
-		if not isSilent then
-			print(`[Luau Language Server] Successfully sent full DataModel info`)
-		end
+		Connected = true
+		ConnectButton.Icon = "rbxassetid://11116536087"
+		warn("[Luau Language Server] Now sending DataModel information.")
```
Barocena, line 193: This warns when any object added/removed, it should only warn on initial connection

plugin/src/init.server.lua
```diff
@@ -1,222 +1,278 @@
...
-local connections = {}
+local DefaultSettingsModule = Instance.new("ModuleScript")
+DefaultSettingsModule.Name = SETTINGS_MODULE_NAME
+ScriptEditorService:UpdateSourceAsync(DefaultSettingsModule, function()
+	return DEFAULT_SOURCE
+end)
+local DefaultSettings: {
+	Port: number,
+	StartAutomatically: boolean,
+	DebugMode: boolean,
```
Barocena, line 83: I dont think anyone other than people contributing to code would need anything more detailed than the current logs

plugin/src/init.server.lua
```diff
@@ -1,222 +1,278 @@
...
+		table.insert(
+			Connections,
+			Service.DescendantAdded:Connect(function(Descendant: Instance)
+				if table.find(CLASSES_EXCEPTION_LIST, Descendant.ClassName) ~= nil then
+					return
+				end
+				--print(`[Luau Language Server] {Descendant.Name} added to {Service.ClassName}`)
+				table.insert(
+					Connections,
+					Descendant.Changed:Connect(function(Property: string)
```
Barocena, line 225: how does Changed perform on large places ? and why we send dmInfo on all the changes when we can just track the name with `GetPropertyChangedSignal("Name")` , we only need name for the dminfo

plugin/src/init.server.lua
```diff
@@ -1,222 +1,278 @@
...
-	-- TODO: we should only send delta info if possible
-	local function descendantChanged(instance: Instance)
-		for _, service in Settings.include do
-			if instance:IsDescendantOf(service) then
-				deferSend()
-				return
+	for _, Service: Instance in Settings.Include do
+		table.insert(
+			Connections,
+			Service.DescendantAdded:Connect(function(Descendant: Instance)
```
Barocena, line 218: it might be expensive to track all the property changes on all the objects

plugin/src/init.server.lua
```diff
@@ -1,222 +1,278 @@
...
+				)
+			end)
+		)
+		for _, Descendant in Service:GetDescendants() do
+			if table.find(CLASSES_EXCEPTION_LIST, Descendant.ClassName) ~= nil then
+				continue
 			end
+			table.insert(
+				Connections,
+				Descendant.Changed:Connect(function(Property: string)
```
Barocena, line 238: `GetPropertyChangedSignal("Name")` here too

Barocena (changes requested):
there are some stuff that needs change and we should benchmark the performance on large place

plugin/src/init.server.lua
```diff
@@ -1,222 +1,278 @@
---!strict
```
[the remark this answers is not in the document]
TaylorsRus: Removed just because off luarc, fine to re-implement

plugin/src/init.server.lua
TaylorsRus: Agreed, this was a left over fragment from when I was planning to do DebugMode in this pr

plugin/src/init.server.lua
TaylorsRus: Addressed in another comment

plugin/src/init.server.lua
TaylorsRus: Addressed in another comment

plugin/src/init.server.lua
TaylorsRus: Addressed in another comment

plugin/src/init.server.lua
TaylorsRus: No performance gripes on medium sized place from me, however can agree that other properties are not neccesary, DescendantAdded handles the Parent, and then Name is the only neccesary field. That should then address potential performance issues with property changes on all descendants, since Name changing will most definitely not cause problems.

Will also be able to remove CLASSES_EXCEPTION_LIST and related functionality, meaning no need to implement one of the future features I mentioned

plugin/src/init.server.lua
```diff
@@ -1,222 +1,289 @@
 --!strict
 local HttpService = game:GetService("HttpService")
-assert(plugin, "This code must run inside of a plugin")
+local RunService = game:GetService("RunService")
+local ScriptEditorService = game:GetService("ScriptEditorService")
+local TestService = game:GetService("TestService")
 
-if game:GetService("RunService"):IsRunning() then
+local IsRunning = RunService:IsRunning()
```
JohnnyMorganz, line 7: I don't really understand the point in putting this into it's own variable, the original seemed fine

plugin/src/init.server.lua
```diff
@@ -1,222 +1,289 @@
...
+	"Toggle Menu",
+	"rbxassetid://11115506617",
+	"Luau Language Server"
+)
+local SettingsButton = Toolbar:CreateButton("Settings", "Open Settings", "rbxassetid://13997395868")
 
-local connected = Instance.new("BoolValue")
-local connections = {}
+local DefaultSettingsModule = Instance.new("ModuleScript")
+DefaultSettingsModule.Name = SETTINGS_MODULE_NAME
```
JohnnyMorganz, line 71: What is the point in inlining this script contents and then creating an instance. Was something wrong with the predefined `DefaultSettings.lua`? This seems like a roundabout way to end up just doing the same thing that was there before.

plugin/src/init.server.lua
```diff
@@ -1,222 +1,289 @@
...
+		)
+		SettingsModule = DefaultSettingsModule:Clone()
+	elseif typeof(Settings.Port) ~= "number" then
+		warn(
+			`[Luau Language Server] Could not load settings: Port is not a number. Defaulting to {DefaultSettings.Port}.`
+		)
+		Settings.Port = DefaultSettings.Port
+	elseif type(Settings.StartAutomatically) ~= "boolean" then
+		warn(
+			`[Luau Language Server] Could not load settings: StartAutomatically is not a boolean. Defaulting to {DefaultSettings.Port}.`
```
JohnnyMorganz, line 114: Should this be `DefaultSettings.StartAutomatically`?

plugin/src/init.server.lua
```diff
@@ -1,222 +1,289 @@
...
+		)
+		Settings.Include = DefaultSettings.Include
+	end
 
-local wasConnected
-SettingsButton.Click:Connect(function()
-	plugin:OpenScript(SettingsModule)
-	wasConnected = connected.Value
-end)
+	return SettingsModule
```
JohnnyMorganz, line 124: We perform edits to `Settings` but end up just returning `SettingsModule`, won't this discard all the edits we made?

plugin/src/init.server.lua
```diff
@@ -1,222 +1,289 @@
...
+local function FilterServices(Child: Instance): boolean
+	return not not table.find(Settings.Include, Child)
 end
 
-local function encodeInstance(instance: Instance, childFilter: ((Instance) -> boolean)?): EncodedInstance
-	local encoded = {}
-	encoded.Name = instance.Name
-	encoded.ClassName = instance.ClassName
-	encoded.Children = {}
+local function EncodeInstance(Instance: Instance, ChildFilter: ((Instance) -> boolean)?): EncodedInstance
```
JohnnyMorganz, line 134: Any particular reason to switch the casing of all these variables?

In particular, `Instance` clashes with the global Instance library, please change this.

plugin/src/init.server.lua
```diff
@@ -1,222 +1,289 @@
...
+	local Encoded = {}
+	Encoded.Name = Instance.Name or "EmptyName"
```
JohnnyMorganz, line 136: Why do we set `or "EmptyName"` and `or "EmptyClassName"` here? AIUI, these should always be set for a given instance. `EmptyClassName` in particular might break the server since it's an invalid class name.

plugin/src/init.server.lua
```diff
@@ -1,222 +1,289 @@
...
-	Name: string,
-	ClassName: string,
-	Children: { EncodedInstance },
-}
+	--// No loss in not running a check here since Luau runs an internal validation for set
+	SettingsModule.Name = SETTINGS_MODULE_NAME
+	SettingsModule.Parent = TestService
+
+	local Settings = require(SettingsModule) :: any
+	print(Settings)
```
JohnnyMorganz, line 101: Should we be printing this?

JohnnyMorganz:
Thanks for taking the time to make this PR!

I've decided to implement fixes to the underlying issues in ea7bea8037aef587476aa3f5692ef5f3635f6c6e and 492f7a43d3db47c509522308fae842105388dd11.

I do not think a complete rewrite of the plugin is needed at this time, as it does not seem to add any value, so I will be closing this PR for now.

If you have any future smaller, specific changes, feel free to open a new one!

Issue #271, opened 2023-01-22
Support doc comments in definition files


JohnnyMorganz:

JohnnyMorganz:
Blocked on https://github.com/Roblox/luau/issues/817

JohnnyMorganz:
We need to provide a way to specify definition files names for this to work correctly.

JohnnyMorganz:
We are still missing location information. Blocked on https://github.com/luau-lang/luau/issues/1137

PR #885, opened 2024-12-29
Require a name for definitions files

JohnnyMorganz:
Makes way for supporting intellisense and documentation comment look up for definitions files

Part of #271 

Supersedes #518


JohnnyMorganz:
There are some broken assumptions on ordering after we switch to an unordered map...

PR #889, opened 2025-01-04
Make `Sound.SoundGroup` optional

BlueStaggo:
Not all sounds have a SoundGroup associated with them. I found out this flaw in one of my scripts I ported to Rojo and analysed in strict mode


JohnnyMorganz (approved):
Thanks!

PR #894, opened 2025-01-09
Load ignore globs and definition files from settings file

### LINKED ISSUE #892: Read ignore globs from settings in luau-lsp analyze
I have my ignore globs defined in my LSP config file and currently you need to re specify the same ignore globs using the `--ignore` flag. This can be inconvenient if you use the same command across many files.

My suggestion is to make `luau-lsp analyze` command read the ignore blobs from the specified settings json file, in addition to the ignore flag.

Stefanuk12:
Closes #892.

Note: I have near to no C++ experience so my solution may not be optimal. However, it does work.


JohnnyMorganz:
Thanks, this is looking good to me!

Although we don't have a test for settings in general right now, do you think you could extract this out into a function and maked it tested in `AnalyzeCli.test.cpp`?

Stefanuk12:
Yeah, I tried to add a couple tests. Not sure if they're how you're supposed to do them but you're free to make edits.

JohnnyMorganz (approved):
Thank you! Sorry for the delay.

I just simplified the test case slightly and added an entry to the changelog

PR #897, opened 2025-01-11
Add server path configuration option to the VSCode extension

cswimr:
This PR adds a configuration value (`luau-lsp.serverPath`) to the VSCode extension that lets you change the path of the Luau LSP server binary. This is useful for platforms like NixOS, where the included binary doesn't work by default. This hasn't been tested yet.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -151,6 +151,11 @@
           "scope": "window",
           "default": "off"
         },
+        "luau-lsp.server.path": {
+          "markdownDescription": "Path to the Luau LSP server binary. If not provided, uses the binary included in the extension.",
+          "type": "string",
+          "default": ""
+        },
         "luau-lsp.ignoreGlobs": {
           "markdownDescription": "Diagnostics will not be reported for any file matching these globs unless the file is currently open",
           "type": "array",
--- a/editors/code/src/extension.ts
+++ b/editors/code/src/extension.ts
@@ -181,11 +181,27 @@ const startLanguageServer = async (context: vscode.ExtensionContext) => {
     }
   }
 
-  const serverBinPath = vscode.Uri.joinPath(
-    context.extensionUri,
-    "bin",
-    os.platform() === "win32" ? "server.exe" : "server",
-  ).fsPath;
+  const serverBinConfig = vscode.workspace
+    .getConfiguration("luau-lsp.server")
+    .get("path", "");
+
+  const uri = vscode.Uri.file(serverBinConfig);
+  let serverBinPath;
+
+  if (await utils.exists(uri)) {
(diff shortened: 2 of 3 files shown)


cswimr:
Update: I've tested this and it works great!

editors/code/package.json
```diff
@@ -151,6 +151,11 @@
           "scope": "window",
           "default": "off"
         },
+        "luau-lsp.serverPath": {
```
JohnnyMorganz, line 154: Can we name this `luau-lsp.server.path`?

It allows us to extend the server configuration with other options in future

```suggestion
        "luau-lsp.server.path": {
```

editors/code/src/extension.ts
```diff
@@ -181,11 +181,16 @@ const startLanguageServer = async (context: vscode.ExtensionContext) => {
...
 
-  const serverBinPath = vscode.Uri.joinPath(
-    context.extensionUri,
-    "bin",
-    os.platform() === "win32" ? "server.exe" : "server",
+  const serverBinConfig = vscode.Uri.file(
+    vscode.workspace.getConfiguration("luau-lsp").get("serverPath", ""),
   ).fsPath;
+  const serverBinPath = serverBinConfig
+    ? serverBinConfig
```
JohnnyMorganz, line 188: Calling `vscode.Uri.file` with an empty string, and then relying on fsPath being undefined, seems odd to me. Maybe we should do something like below:

```suggestion
  const serverBinConfig = vscode.workspace.getConfiguration("luau-lsp").get("serverPath");
  const serverBinPath = serverBinConfig
    ? vscode.Uri.file(serverBinConfig).fsPath
```

Also, similar to how we load definitions files, it might make sense to do an `exists` check on the provided binary to check if it is actually there, and emit an error if it isn't.

JohnnyMorganz (changes requested):

editors/code/src/extension.ts
cswimr: I'll take a look at doing that `exists` check, thanks for the review!

cswimr:
Marking this as a draft until I clean up the nixos files

cswimr:
Accidentally did a merge, so I'm gonna condense my changes into a single commit and force push

JohnnyMorganz (approved):
Thanks! Sorry for the delay. Can we update the changelog with this?

cswimr:
> Thanks! Sorry for the delay. Can we update the changelog with this?

No problem, and yeah, I'll commit that real quick. Thanks for the review!

cswimr:
oops that merged, sigh

JohnnyMorganz (approved):
Thank you!

PR #908, opened 2025-01-23
Support pretty diagnostic messages in CLI #495

### LINKED ISSUE #495: Support pretty diagnostic messages in CLI
Would be nice to be able to report something in a similar fashion to Rust's https://github.com/brendanzab/codespan in the analyze tool

There is https://github.com/Excse/pretty_diagnostics already that is somewhat similar. Or maybe it doesn't take too much effort to port the codespan library

danaetroupe:
Closes #495: **You can now use the --formatter prettier flag in the CLI to get prettier errors**

NOTES: 
- Characters may not show properly in certain outdated terminals
- Multi-line errors not supported
- See issue for more details about implementation 

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ trace.json
 
 # Python venvs
 .venv
+
+# Test File
+discard.luau
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,72 +5,92 @@
 {
   "version": "0.2.0",
   "configurations": [
+    // {
+    //   "name": "Run Extension Standalone",
+    //   "type": "extensionHost",
+    //   "request": "launch",
+    //   "args": [
+    //     "--disable-extensions",
+    //     "--extensionDevelopmentPath=${workspaceFolder}/editors/code"
+    //   ],
+    //   "outFiles": ["${workspaceFolder}/editors/code/dist/**/*.js"],
+    //   "preLaunchTask": "npm: compile",
+    //   "env": {
+    //     "LUAU_LSP_SERVER_PATH": "${workspaceFolder}/${input:binaryPath}/luau-lsp"
+    //   },
+    //   "windows": {
+    //     "env": {
+    //       "LUAU_LSP_SERVER_PATH": "${workspaceFolder}/${input:binaryPath}/luau-lsp.exe"
+    //     }
+    //   }
+    // },
+    // {
+    //   "name": "Run Extension",
+    //   "type": "extensionHost",
+    //   "request": "launch",
+    //   "args": [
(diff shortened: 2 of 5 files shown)


src/include/Analyze/Reporter.hpp
JohnnyMorganz, line 1: Can you move this into the `extern` directory? That is where we keep our third party code.

Also adding a README pointing to the repository you got this from would be useful for future reference

.gitignore
JohnnyMorganz, line 1: Please revert the changes to this file once complete

.vscode/launch.json
JohnnyMorganz, line 1: Please revert the changes to this file once complete

src/AnalyzeCli.cpp
```diff
@@ -54,6 +56,15 @@ static void report(ReportFormat format, const char* name, const Luau::Location&
         // Note: GNU end column is inclusive but our end column is exclusive
         fprintf(stderr, "%s:%d.%d-%d.%d: %s: %s\n", name, loc.begin.line + 1, loc.begin.column + 1, loc.end.line + 1, loc.end.column, type, message);
         break;
+
+    case ReportFormat::Prettier:
+        // Use Reporter.hpp to create new reporter error
+        auto file = new reporter::SimpleFile(name);
+        auto err = reporter::Error(
+                type, message,
```
JohnnyMorganz, line 64: I wonder if the top line should include the error message, rather than just `TypeError` - i think that should go as part of the code.

JohnnyMorganz (commented):
Sorry for the delay here.

This is looking pretty neat, and surprisingly simple implementation too, I guess everything we need is already there in the 3rd party repo :)

I'm actually curious, maybe we should enable this as the default output. I think it would be useful for most users.

Anything in particular you think this is still missing?

danaetroupe:
@JohnnyMorganz There are some terminal incompatibilities so I'd like to see what the results are on different machines. I want to look into other character sets. In my terminal, I had to manually allow it to show certain characters so I wonder if there is a way to detect that. Also, I think ultimately being able to group errors by file will be important. I can come back and take a look at those changes in depth but I think that should be a feature before enabling as default. If you are analyzing an entire project at once I could see this formatting becoming inefficient.

danaetroupe:
@JohnnyMorganz I got busy with school, but I want to take a look at this again. Let me know if there's anything else you want from this.

JohnnyMorganz:
Sorry @danaetroupe, I should've also followed up on this too.

I'd be happy to re-review the initial implementation you had and merge it in so we can start building improvements on it

PR #910, opened 2025-01-25
fix: return value in Roblox platform notification handler

lopi-py:
https://github.com/JohnnyMorganz/luau-lsp/blob/a82ebc1823fa74fa358c8401ff2c7adcf0ea7aa7/src/LanguageServer.cpp#L342-L348
We don't return `true` so we still hit "unknown notification method" even if handled

### DIFF
--- a/src/LanguageServer.cpp
+++ b/src/LanguageServer.cpp
@@ -339,13 +339,15 @@ void LanguageServer::onNotification(const std::string& method, std::optional<jso
     }
     else
     {
+        bool handled = false;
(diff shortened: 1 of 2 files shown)


JohnnyMorganz:
I think we need to be cautious here since this early return will stop multiple workspace folders from working. Before, it would update the plugin data for all workspace folders, but after your change it will stop after the first workspace folder, and not update the others

lopi-py:
Indeed, I have updated it

JohnnyMorganz (approved):
Thanks!

PR #912, opened 2025-01-26
Fix Types for Faces and Axes datatypes

### LINKED ISSUE #911: Axes.new and Faces.new are incorrectly typed
Both functions should take in multiple arguments.

Axes.new should take multiple `Enum.NormalId | Enum.Axis`'s while Faces.new should take multiple `Enum.NormalId`'s.

Repro:
```luau
local axes = Axes.new(Enum.Axis.X, Enum.Axis.Y, Enum.Axis.Z) -- lsp error: function only takes one argument
local faces = Faces.new(Enum.NormalId.Top, Enum.NormalId.Bottom) -- lsp error: function only takes one argument
```

Barocena:
Closes #911 

This PR fixes the types of Faces and Axes datatypes.

### DIFF
--- a/scripts/DataTypes.json
+++ b/scripts/DataTypes.json
@@ -689,9 +689,12 @@
                     "Parameters": [
                         {
                             "Name": "axes",
-                            "Type": {
-                                "Name": "Tuple"
-                            }
+                            "Type":{
+                                "Variadic":{
+                                    "Category": "Enum",
+                                    "Name": "Axis | EnumNormalId"
+                                }
+                        }
                         }
                     ],
                     "ReturnType": {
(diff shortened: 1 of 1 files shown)


scripts/DataTypes.json
```diff
@@ -689,9 +689,12 @@
                     "Parameters": [
                         {
                             "Name": "axes",
-                            "Type": {
-                                "Name": "Tuple"
-                            }
+                            "Type":{
+                                "Variadic":{
+                                    "Category": "Enum",
+                                    "Name": "Axis | EnumNormalId"
```
JohnnyMorganz, line 695: This union within an enum seems weird to me, does it actually generate the expected union after running the dump script?

Barocena: yes, Axis end up as EnumAxis but it wasn't changing the normalid since it is union type
it generates  `new: ((...EnumAxis | EnumNormalId) -> Axes)` for the Axes

JohnnyMorganz: Cool, sounds good

JohnnyMorganz (approved):

PR #915, opened 2025-01-27
feat: add require path in labelDetails.description

### LINKED ISSUE #891: `labelDetails.description` require path
a9e339d51cf08adb4816e1a139a011fba19c001a improved the detail section of the suggested require, would be nice to also have it in `labelDetails.description`

typescript lsp example:
image

lopi-py:
Closes #891

image

Not sure if we should also show `Auto-import` as label detail for services


JohnnyMorganz (approved):
Thanks!

PR #917, opened 2025-01-29
Insert a newline for empty single-line documentation comments

### LINKED ISSUE #905: Single-line comments don't add newlines
```luau
--- @class MyClass
--- 
--- A sample class.
local MyClass = {}
MyClass.__index = MyClass
```

Image

```luau
--[=[
	@class MyClass

	A sample class.
]=]
local MyClass = {}
MyClass.__index = MyClass
```

Image

maffeus:
This PR makes empty single-line documentation comments act as newlines. 
This change fixes #905 and aligns single-line comments with multi-line comments, making them essentially equivalent as far as I can tell.

```luau
--- @class MyClass
---
--- A sample class.
local MyClass = {}
MyClass.__index = MyClass
```

image

```luau
--[=[
	@class MyClass

	A sample class.
]=]
local MyClass = {}
MyClass.__index = MyClass
```

image


JohnnyMorganz:
Thanks for this! Can you also add a test case to `Documentation.test.cpp`?

JohnnyMorganz:
Whoops, wrong button

maffeus:
I'm not sure if this is the best way to add a test like this, because I feel like this is a much more complicated and nuanced test. Probably what would need to happen is every use of multiline comments (or vice versa) in `Documentation.test.cpp` should probably be replaced with a use of both (?) and check if their results are equal, but that just seems terrible to implement, and I'm not going to change the entire codebase for just that. 😓

Ideally, the end goal should probably _always_ be congruence between multiline and single-line doc comments, such that someone could replace the `--- ` with tabs, and add the `--[=[` prefix and `]=]` suffix and be left with the same rendered result no matter what. I'm just not sure how you would test such a thing without just testing both multiline and single-line doc comments for every test case.

JohnnyMorganz:
This test makes sense to me! Although it looks like it is failing?

I also fixed a compilation error and added a simpler test case. I'm not sure what is the expected end result of the test you added though - it looks like the number of newlines between the singleline and multiline definitions are different?

Could you also add this to the changelog?

maffeus:
The idea was just to check if single-line documentation comments were equivalent to multi-line documentation comments but honestly the test you added is better and covers what we actually needed so I think I'll just go ahead and remove the one I made, C++ is cruel 😭

JohnnyMorganz (approved):
Clear, thank you!

PR #927, opened 2025-02-14
Set PluginToolbarButton ClickableWhenViewportHidden to true

RuizuKun-Dev:
made both PluginToolbarButton ClickableWhenViewportHidden = true

Currently when I'm editing Plugin Settings and want to reconnect I'd have to exit out of the Script which is annoying if I want to try re-connecting

I don't see a reason why ClickableWhenViewportHidden should be kept as false

image


JohnnyMorganz (approved):
Thank you!

PR #940, opened 2025-02-17
Fix Vector2:Angle return type

### LINKED ISSUE #939: Roblox `Vector2:Angles()` has incorrect return type!
`Vector2:Angles()` should return a number _(as described in the [docs](https://create.roblox.com/docs/reference/engine/datatypes/Vector2#Angle))_ **_however_** **`Luau LSP`** says that is should return a Vector2!

Image

Barocena:
Closes #939

### DIFF
--- a/scripts/DataTypes.json
+++ b/scripts/DataTypes.json
@@ -2787,7 +2787,7 @@
                         }
                     ],
                     "ReturnType": {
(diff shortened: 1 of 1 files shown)


JohnnyMorganz (approved):
Thanks!

PR #956, opened 2025-03-05
Attach documentation comments to type alias definitions

### LINKED ISSUE #953: Attach comments to intersections (or potentially on alias definitions?)
I'm honestly not sure if this is the right title for this issue.

```luau
--- An animal
type Animal = {
    species: string,
}

--- A cat
type Cat = Animal & {
    meow: (self: Cat) -> "Meow!",
}
```

In this example, when hovering over the `Animal` type, you can see the comment `An animal` as expected, but when hovering over the `Cat` type, you don't see anything.

A common pattern to allow for recursive type definitions is to create a `Data` type and a `Methods` type, like this:

```luau
type OptionData<T> = {
    value: T?,
}

type OptionMethods = {
    map: <T, U>(
        self: OptionData<T>,
        f: (T) -> U
    ) -> OptionData<U> & OptionMethods,
}

export type Option<T> = OptionData<T> & OptionMethods
```

But it's a bit annoying that you can't have comments on the `Option<T>` type in this case.

m4fh:
In #953, we discussed the possibility of attaching documentation comments to intersections. Since that isn't currently possible due to how Luau works, you mentioned that it might be possible to attach comments to type alias definitions instead.

I implemented this by adding a new `getDocumentationForAstNode` function, which runs if `getDocumentationForType` doesn’t return anything. I’m not sure if this is the best approach, but it works. 

Now code like this is actually working, which, in my opinion, is a __huge__ improvement to library documentation:

```luau
--- The metre (or meter in [US spelling]; symbol: m) is the [base unit] of [length] 
--- in the [International System of Units] (SI)
---
--- [US spelling]: https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-re,_-er
--- [base unit]: https://en.wikipedia.org/wiki/SI_base_unit
--- [length]: https://en.wikipedia.org/wiki/Length
--- [International System of Units]: https://en.wikipedia.org/wiki/International_System_of_Units
export type Meters = number;
```

Fixes #953


m4fh:
On a side note, if this PR gets merged, I propose that this:

```luau
--- A comment
type Original = any

type Alias = Original
```

should be _explicitly_ equivalent to:

```luau
--- A comment
type Original = any

--- A comment
type Alias = Original
```

Previously, this only worked when the type alias didn’t have type parameters. I believe this was because the docs on `Original` were attached to the table or function. So, when you had an alias like `type Alias<T> = Original<T>`, it was treated as a distinct type and lost the original documentation.

Now that type alias definitions can have docs attached directly, it makes sense to extend this behavior.

I suggest that if `Alias` has no docs, meaning it has no comments at all (even an empty comment could cancel this effect),   
and `type Alias<T...> = Original<T...>`, then we should copy the comments from `Original` into `Alias`.  

This would allow developers to re-export type aliases without losing documentation, like this:

```luau
-- internalTypes.luau

--- A comment
export type InternalType<T> = {
    parameter: T,
}
```

```luau
-- publicTypes.luau

local internalTypes = require("internalTypes")

export type Type<T> = internalTypes.InternalType<T>
```

JohnnyMorganz (commented):
This looks pretty cool, thank you!

Could you please add some tests for the new method in DocumentationParser and also showing it working correctly in some hover tests?

Also, does this make sense to add for other operations too, such as autocomplete or signature help? Autocomplete might be a bit more challenging since I don't think we have the source AST node available anywhere

m4fh:
> Could you please add some tests for the new method in DocumentationParser and also showing it working correctly in some hover tests?

Sure, I'll do that right now.

> Also, does this make sense to add for other operations too, such as autocomplete or signature help? Autocomplete might be a bit more challenging since I don't think we have the source AST node available anywhere

I tried getting autocomplete to work (as far as I can tell it's the only other use of `getDocumentationForType`?) but I didn't get very far.

JohnnyMorganz:
> On a side note, if this PR gets merged, I propose that this:

I think this seems reasonable to me, but out of scope for this PR.

My only concern is it becoming challenging to figure out the correct source and flow of doc comments, but this particular case (of reexporting types) sounds reasonable and useful.

As an adjacent point (also out of scope for this PR itself), I think this PR also makes way for support doc comments on general variable definitions too:
```
--- My special constant
local MY_CONSTANT = 500
```
But might only be possible for local variables used within a file, and not variables that make their way into the exported interface (or maybe it is? We'll have to see!)

m4fh:
> My only concern is it becoming challenging to figure out the correct source and flow of doc comments, but this particular case (of reexporting types) sounds reasonable and useful.

I've seen this pattern a lot in codebases, people usually do it to avoid recursive require issues. You have a types.luau, and every submodule requires it instead of requiring the main module (the one the user would actually require). But then you run into the awkward situation of having to choose between:

- Keeping documentation only in types.luau,
- Keeping documentation only in the main module,
- Repeating yourself, or
- Having the user require types.luau directly (which requires two require() calls for a single module), and so on.

JohnnyMorganz:
I was just wondering if this does actually completely fix #953 as expected?

This will add the documentation comment when hovering over the type reference itself, but it doesn't pass over to the variables that have that type

image

JohnnyMorganz:
Also added a test case for references which I think was missing, as the tests only covered definitions

m4fh:
> I was just wondering if this does actually completely fix #953 as expected?
> 
> This will add the documentation comment when hovering over the type reference itself, but it doesn't pass over to the variables that have that type
> 
> image

In #953, I was thinking about documentation on types (as in, if you're making a library, you want your types to be properly documented so that people don’t have to look up docs on an external site just to understand how your library works). So, I guess my original issue does get fixed with this PR

But thinking about the overarching goal of the PR #956, I can see how it would be useful (and important) to make it consistent with the other ways you can attach comments to types. I could fix just the example you sent pretty easily, but I think I'll have to take a look at the `hover` method to see if I can apply everything you already do for `getDocumentationForType` to `getDocumentationForAstNode`, so everything is nice and consistent

JohnnyMorganz:
All good, just wanted to confirm. The current work is still useful regardless. Looks good to me!

We can work on extending it further in follow up PRs

JohnnyMorganz (approved):

PR #957, opened 2025-03-05
Re-alias ContentId as string, remove Content alias.

MaximumADHD:
`ContentId` is the new name for what was previously the `Content` type, before the actual Content type was introduced.
I can't auto-complete string methods on ContentId properties at the moment because of the `any` type. It should be `string`.


JohnnyMorganz (approved):
Thanks!

PR #964, opened 2025-03-08
Enable HTML rendering in language client options

m4fh:
This PR enables HTML support by adding `supportHtml: true` to the language client options and removing the `convertHtmlToMarkdown` function, which previously just removed the HTML tags from documentation.

I assume other language clients also support HTML (perhaps with a similar option that just needs to be toggled on?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Enabled HTML support by setting `supportHtml: true` in the language client options. ([#964](https://github.com/JohnnyMorganz/luau-lsp/pull/964))
+- Enabled HTML support by setting `supportHtml: true` in the language client options ([#964](https://github.com/JohnnyMorganz/luau-lsp/pull/964))
 - Documentation comments now attach to type alias definitions ([#956](https://github.com/JohnnyMorganz/luau-lsp/pull/956))
 
 ### Changed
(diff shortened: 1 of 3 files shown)


JohnnyMorganz:
Can you show an example of before and after this change?

Also maybe we should reference `allowedTags` from the language client to filter the tags we remove: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markdownClientCapabilities. This means it will still apply for language clients that do not support HTML in markdown

m4fh:
> Can you show an example of before and after this change?

Here are some quick examples:

```luau
--- <img width=48 src="https://avatars.githubusercontent.com/u/175355178?v=4"/>
--- <hr/>
--- My GitHub account
local my_github_account = {}

--- If the player is currently pressing <kbd>E</kbd>
local function is_pressing_e() end
```

This is how they look like with this PR:

image
image

m4fh:
> Also maybe we should reference `allowedTags` from the language client to filter the tags we remove: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markdownClientCapabilities. This means it will still apply for language clients that do not support HTML in markdown

That is a good idea for sure, but I think it could be part of a different PR. My reasoning for that is that the old `convertHtmlToMarkdown` function didn't really work, for example, escaped characters (`&amp;`) just didn't get converted at all. So, the language clients are kind of (as far as I can tell) forced to handle HTML on their end. With VS Code, for example, the reason these weren't ever noticed is that VS Code automatically handles that for us (by stripping HTML themselves if `supportHtml` isn't set to true).

So, this update shouldn't really break any current language client code (although it'd probably be interesting to look into those codebases a bit further, which I must admit I haven't done).

JohnnyMorganz (approved):
That reasoning makes sense to me, thanks!

It would be good to check in with neovim / zed to see if they will struggle. But happy to move forward with this and we can revert if necessary.

Can you update the changelog?

PR #976, opened 2025-03-20
feat: Add virtual nevermore loader support

Quenty:


JohnnyMorganz:
Closing this as I expect https://github.com/JohnnyMorganz/luau-lsp/pull/1088 will be sufficient to support this need. Happy to extend plugins mechanism further if there's anything missing

PR #1088, opened 2025-05-31
Plugins

JohnnyMorganz:


JohnnyMorganz:
Sample plugin implementation of #976 in its current state:

```luau
type SourcemapNode = {
	name: string,
	className: string,
	filePaths: { string }?,
	children: SourcemapNode
}

local function startsWith(str: string, prefix: string)
	return str:sub(1, #prefix) == prefix
end

local function endsWith(str: string, suffix: string)
	return str:sub(-#suffix) == suffix
end

local REQUIRE_PATTERN = [[require%("(%w+)"%)]]

local root = lsp.workspace.getRootUri()
local sourcemapContents = lsp.fs.readFile(root:joinPath("sourcemap.json"))
local sourcemap = lsp.json.deserialize(sourcemapContents) :: SourcemapNode

local function getScriptFilePath(node: SourcemapNode): string?
	if not node.filePaths then
		return nil
	end

	for _, path in node.filePaths do
		if endsWith(path, ".luau") or endsWith(path, ".lua") then
			return path
		end
	end

	return nil
end

-- Populate module names to file paths map
local moduleNamesToFilePaths = {}
local queue = {sourcemap}
while #queue > 0 do
	local node = table.remove(queue, 1)
	assert(node ~= nil)

	local filepath = getScriptFilePath(node)
	if filepath then
		moduleNamesToFilePaths[node.name] = root:joinPath(filepath)
	end

	if node.children then
		for _, child in node.children do
			table.insert(queue, child)
		end
	end
end

return {
	transformSource = function(source: string, context: PluginContext): { TextEdit }?
		local edits = {}
		local lines = string.split(source, '\n')

		for i, lineContents in lines do
			local lineNumber = i - 1

			-- Comment out the require override to prevent it from impacting intellisense
			if startsWith(lineContents, "local require = ") then
				table.insert(edits, {
					range = {
						start = { line = lineNumber, column = 0 },
						["end"] = { line = lineNumber, column = 0},
					},
					newText = "-- "
				})
			else
				local startIndex, endIndex = string.find(lineContents, REQUIRE_PATTERN)
				if startIndex then
					local REQUIRE_START_LEN = string.len('require("')
					local REQUIRE_END_LEN = string.len('")')

					local nameStartIndex = startIndex + REQUIRE_START_LEN
					local nameEndIndex = endIndex - REQUIRE_END_LEN

					local requireName = lineContents:sub(nameStartIndex, nameEndIndex)

					if moduleNamesToFilePaths[requireName] then
						table.insert(edits, {
							range = {
								-- column is 0 indexed
								start = { line = lineNumber, column = nameStartIndex - 1 },
								["end"] = { line = lineNumber, column = nameEndIndex },
							},
							newText = moduleNamesToFilePaths[requireName].fsPath
						})
					end
				end
			end
		end

		return edits
	end,
}
```

PR #1106, opened 2025-06-11
Redefine RaycastResult to have generic type (with BasePart default)

MaximumADHD:
This is a correctly redone attempt to do what I attempted last year. I've tested it in my own environment and can confirm it works as intended.

The only side effect I've observed is that it breaks type refinement for `typeof(RaycastResult)`, but I reckon this might be acceptable since a similar change was made to `RBXScriptSignal`, and even Roblox Studio defines the type as a dictionary. In an ideal world, class declarations would support generics... but this'll have to do for now.


JohnnyMorganz (approved):
Thanks!

PR #1107, opened 2025-06-12
Add FromValue and FromName Methods to Enum

Barocena:
This PR adds FromValue and FromName Methods to Enum.

Return type is `any` since we can't return the actual enum (`Enum.AssetType:FromValue()` returns `Enum.Assettype?` in studio)

### DIFF
--- a/scripts/dumpRobloxTypes.py
+++ b/scripts/dumpRobloxTypes.py
@@ -366,6 +366,8 @@ type AdReward = any
 
 declare class Enum
     function GetEnumItems(self): { any }
+    function FromValue(self,Number: number): any
+    function FromName(self,Name: string): any
 end
 
 declare class EnumItem


JohnnyMorganz (approved):
> Return type is `any` since we can't return the actual enum (`Enum.AssetType:FromValue()` returns `Enum.Assettype?` in studio)

We could probably update our Enum generation logic to put an override in each of the enum classes. i.e. generate:

```luau
declare class EnumAssetType_INTERNAL extends Enum
    Animation: EnumAssetType
    Audio: EnumAssetType
    ...
    function GetEnumItems(self): { EnumAssetType }
    function FromName(self, Name: string): EnumAssetType?
    function FromValue(self, Value: number): EnumAssetType?
end
```

PR #1110, opened 2025-06-13
feat: add server info to lsp initialize results

AleksandrSl:
Adds information about LSP to the initialization result, so it can be used by lsp integrations (e.g. JB IDE plugin)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,11 @@ option(LSP_BUILD_ASAN "Build with ASAN" OFF)
 option(LSP_STATIC_CRT "Link with the static CRT (/MT)" OFF)
 option(LSP_BUILD_WITH_SENTRY "Build with Sentry (crash reporting)" OFF)
 
+set(LSP_VERSION "1.49.1")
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -6,6 +6,11 @@ option(LSP_BUILD_ASAN "Build with ASAN" OFF)
 option(LSP_STATIC_CRT "Link with the static CRT (/MT)" OFF)
 option(LSP_BUILD_WITH_SENTRY "Build with Sentry (crash reporting)" OFF)
 
+set(LSP_VERSION "1.49.1")
+set(LSP_NAME "Luau Language Server")
```
AleksandrSl, line 10: Feel free to change this to `luau-lsp` or whatever name your prefer. It can be a constant inside the `src/LanguageServer.cpp`, but I thought that keeping these two together makes sense

AleksandrSl:
I checked with a local build that I now see the version in my plugin 
Screenshot 2025-06-14 223937

JohnnyMorganz (approved):
Thanks!

PR #1113, opened 2025-06-15
Improve Enum Type Returns

Barocena:
This PR changes the return type of methods on enums from any to actual enum


JohnnyMorganz (approved):
Thanks!

PR #1181, opened 2025-08-17
Fix for CFrame.fromEulerAngles constructor

itred12:
Fixed the "order" variable in CFrame.fromEulerAngles being incorrectly marked as required by the typechecker


JohnnyMorganz (approved):
Thanks!

Issue #967, opened 2025-03-11
Make it clearer that the studio plugin only works with a sourcemap


JohnnyMorganz:
Plugin is purely additive. We should emit a warning if we don't have a sourcemap.

Or maybe the plugin should start working independently?

staurosKaragiannis:
I spent hours trying to figure out how to get this working. I just realised that i can't use the plugin without Rojo. 

I hope the plugin starts working independently, why isn't that the default?

A warning would have saved the headache

JohnnyMorganz:
Sorry about that, the plugin does have a couple rough edges still.

The reason running independently is currently not the default is because the Studio Plugin does not provide any information about how a script maps to a path on your filesystem. Right now we still need the source map to tell this with its `filePaths` information. Originally, my thinking was that the plugin is not useful without that (since all your requires would still not resolve, you'd get red squiggly lines everywhere, and the main assets on the file system is typically scripts). Right now, the Studio Plugin is only useful to provide other DataModel information.

Hopefully this changes in the future as Roblox's Script Sync system matures (since that is independent of Rojo). I believe there are plans to get the `filePaths` information available in plugins, and then we can support that properly.

JohnnyMorganz:
https://github.com/JohnnyMorganz/luau-lsp/pull/1216 now appropriately handles the results from the Studio Plugin even if there is no sourcemap

PR #1216, opened 2025-09-30
Feature: Support DM info from companion plugin even if no sourcemap.json exists

zovits:
The info from the companion plugin is currently unused if there's no `sourcemap.json` file to combine it with.

We can use a default sourcemap of an empty DataModel (in memory, not writing a file) so that you can still utilize the information from the plugin when no sourcemap file is found.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,17 +6,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## [1.55.0] - 2025-10-19
+
 ### Added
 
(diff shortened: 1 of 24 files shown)


JohnnyMorganz:
Looks good, thank you! This will also resolve https://github.com/JohnnyMorganz/luau-lsp/issues/967 in an even better way.

Do you think you could get a test added for this in `Sourcemap.test.cpp`? No worries if it's too painful, the structure might not be in a setup that is nicely testable, so please don't spend too long on trying to get it working. I can look into it later.

JohnnyMorganz (approved):
Thank you!

Issue #1253, opened 2025-11-13
Incorrect default value for 'types.definitionFiles' setting in vscode extension


sircfenner:
With the change to how the definitionFiles setting works in the vscode extension, the 'default' field was not updated from an empty array to an empty object. This creates a type mismatch issue in the settings menu as well as an incorrect return value when using the vscode extension api to retrieve the current or fallback value for a setting.

PR #1254, opened 2025-11-13
Fix default value for 'types.definitionFiles' setting in vscode extension

sircfenner:
Fixes https://github.com/JohnnyMorganz/luau-lsp/issues/1253


JohnnyMorganz:
We're missing a changelog entry for this, unfortunately I can't add it on mobile. But I'll get it merged anyways so that I can put the fix out in a release.

PR #1257, opened 2025-11-18
Update Luau to release 0.700

aatxe:
The auto-bot for syncing to #1248 worked just fine, but we had some API changes for propagating limits around. This PR gives the infrastructure to implement supporting limits and cancellation tokens in the right places, but as-is doesn't implement any actual cancellation support. I cannot seem to find any indication that luau-lsp interacts with `TypeCheckLimits` today, so I added a constructed one to workspaces and that's the basis by which it gets propagated. This should give enough of a basis for follow-up work to actually implement timeouts and cancellation.


JohnnyMorganz (approved):
Thank you for unblocking this!

PR #1278, opened 2025-12-03
Activate VSC extension if the opened workspace contains Luau

zovits:
Instead of waiting for the user to open a Luau file, we can activate when they open a workspace containing Luau files. 

This has two key impacts:
- Opening a Luau file will now give you the LSP features quicker instead of waiting for the LSP to start up
- Connecting the Studio plugin no longer requires you to open a Luau file (to launch the localhost server for it) first

If they open a workspace that does not contain Luau, the extension will only activate if they then create and open a Luau file, same as the current behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- VSCode: the extension will now automatically activate when your opened folder contains any `.luau` files, not just when you explicitly open a Luau file. This allows the Studio Plugin to start up and connect quicker ([#1278](https://github.com/JohnnyMorganz/luau-lsp/pull/1278))
+
 ## [1.59.0] - 2025-12-28
 
 ### Added
@@ -1806,6 +1810,7 @@ local y = tbl.data -- Should give "This is some special information"
 ### Added
 
 - Added configuration options to enable certain Language Server features. By default, they are all enabled:
+
   - `luau-lsp.completion.enabled`: Autocomplete
   - `luau-lsp.hover.enabled`: Hover
   - `luau-lsp.signatureHelp.enabled`: Signature Help


checkraisefold:
Does vscode actually do this check fast or does this incur a performance penalty from glob scanning the files in the project? Also, what about checking for a `.luaurc` file either instead or also? That's usually a nearly 100% positive indicator that there's luau files

I also kind of don't like activating when `.lua` files are found; I also have sumneko's Lua extension installed for the rare times I do regular Lua development and its kind of nice being able to just change the language association in the workspace instead of the extension being disabled. People should just rename their `.lua` files to `.luau` at this point, it's a different language and the luau extension has been standard for a while.

Does the `workspaceContains` selector also infer the old logic, where if I have no workspace open and just open a single `.luau` file the LSP will still start? If not, you should probably add the `onLanguage:luau` back.

zovits:
> does this incur a performance penalty

Negligible impact, thankfully.

> checking for a .luaurc file

Good idea! I don't think it will really have an impact though, since if there's a .luaurc file then there are surely .luau file present as well.

> don't like activating when .lua files are found

Fair, but the .luau extension is still picking up adoption slowly. I'll defer to @JohnnyMorganz to decide whether the extension should boot on Lua workspaces.

> add the onLanguage:luau back

No need, that activation will still happen. From the VSC docs: "Beginning with VS Code 1.74.0, languages contributed by your extension do not require a corresponding onLanguage activation event declaration for your extension to be activated."

checkraisefold:
> No need, that activation will still happen. From the VSC docs: "Beginning with VS Code 1.74.0, languages contributed by your extension do not require a corresponding onLanguage activation event declaration for your extension to be activated."

Neat, that's convenient! This looks fine with me then once a decision is made for `.lua` files.

zovits:
> The lazy initialization infra in the server is set up in such a way that when we should've always performed initialization by the time we need to access the workspace.

Requests to the LSP don't get processed until the workspace is ready. In #1295, the `$/plugin/getFilePaths` request therefore returns a 500 until after a Luau text document is opened. I might be able to work around that by making the extension handle gathering the file paths without requesting it from the LSP.

zovits:
Ah, I can use `vscode.workspace.findFiles` in the extension instead of depending on LSP initialization. I can undo those changes here then.

JohnnyMorganz:
> From the VSC docs: "Beginning with VS Code 1.74.0, languages contributed by your extension do not require a corresponding onLanguage activation event declaration for your extension to be activated."

Our minimum VSCode version is still 1.67.0. We should bump it to 1.74.0 if we expect to rely on this behaviour

> Fair, but the .luau extension is still picking up adoption slowly. I'll defer to @JohnnyMorganz to decide whether the extension should boot on Lua workspaces.

Sorry, I missed this one. What do you think about starting with `.luau` only and expanding to `.lua` if it seems necessary? I'd prefer new features to rely on only `.luau` as much as possible. Does Script Sync generate `.luau` by default?

zovits:
> Our minimum VSCode version is still 1.67.0. We should bump it to 1.74.0 if we expect to rely on this behaviour

I put the explicit onLanguage trigger back to maintain support of older versions.

> Sorry, I missed this one. What do you think about starting with `.luau` only and expanding to `.lua` if it seems necessary? I'd prefer new features to rely on only `.luau` as much as possible. Does Script Sync generate `.luau` by default?

ScriptSync uses Luau by default but can be changed to use Lua instead. We can start with Luau only for now.

PR #1279, opened 2025-12-04
Make plugin installation easy and discoverable

zovits:
Right now, the companion plugin is only mentioned in one line of the readme and the related extension settings. This makes it hard to find and adopt.

To improve this flow, we add a new command that enables the relevant setting, starts the localhost server, and sends the user to the Creator Store page to install the plugin for Studio. When relevant, notifications will include command as a button option.

image


JohnnyMorganz (approved):
Thanks!

PR #1282, opened 2025-12-05
Sync to Luau 0.702, remove deprecated usages

checkraisefold:
tin

deprecated functions were removed entirely and flag was removed as well, so this fixes the compilation error caused. eq sat was also removed, so simplifier has been removed from `Luau::Subtyping` construction.


JohnnyMorganz (approved):
Thanks!

PR #1286, opened 2025-12-06
fix: correct variable in clearDisabledGlobals for ExternType

zoon:
- When handling ExternType globals, the code incorrectly used ttv->props.erase() instead of ctv->props.erase(), causing a null pointer dereference. This was a copy-paste error from the TableType branch above.
Kudos to GCC's `-Wnonnull`

### DIFF
--- a/src/include/Analyze/CliClient.hpp
+++ b/src/include/Analyze/CliClient.hpp
@@ -30,7 +30,7 @@ struct CliClient : public BaseClient
 
     void sendLogMessage(const lsp::MessageType& type, const std::string& message) const override
     {
-        std::cerr << "[" << getMessageTypeString(type) << "] " << message;
+        std::cerr << "[" << getMessageTypeString(type) << "] " << message << "\n";
     }
 
     // In the CLI, this is only used for config errors right now


JohnnyMorganz:
Thanks! Can you add a test for this in Definitions.test.cpp?

JohnnyMorganz:
Note that the definitions loaded for tests is a very limited subset and "BindToClose" already doesn't exist in it https://github.com/JohnnyMorganz/luau-lsp/blob/main/tests/testdata/standard_definitions.d.luau

Maybe try a different option, like Instance.ClearAllChildren

zoon:
https://github.com/JohnnyMorganz/luau-lsp/blob/14991912a4d458bd2f5e8367263a2fd5ee1b01c6/src/Workspace.cpp#L434-L442
```json
{"message":"disabling globals: could not find method - Instance.ClearAllChildren","type":2}
```
`Instance` class is not a global binding, global `Instance` is `TableType` (another path) without `ClearAllChildren`. To trigger this code path (L439) we need a **global** `ExternType` with a direct method (or field) to disable. There is no suitable candidate in definitions right now.

PR #1290, opened 2025-12-12
fix: enable automatic imports server side

lopi-py:
9c8d58ea2156d608dbe496c675d327f2f26acb9e enabled auto-imports by default, but it's client side and other clients like neovim don't send the whole settings schema so it was still disabled by default


JohnnyMorganz (approved):
Thanks!

PR #1298, opened 2025-12-20
feat: convert quotes to backticks on `{` inside strings

lopi-py:
Add `textDocument/onTypeFormatting` support

Automatically convert `"` or `'` quotes to a backtick ` if the user typed an opening bracket within the string. AFAIK this is an opt-in feature already in VSCode but I'm not sure how to enable or test it there.

https://github.com/user-attachments/assets/ee26c11f-4b13-4374-8751-bc58acc9cdc4

---

Can this potentially replace the implementation of `autocompleteEnd`? it seems to relly too much on specific vscode behavior currently

### DIFF
--- a/tests/OnTypeFormatting.test.cpp
+++ b/tests/OnTypeFormatting.test.cpp
@@ -17,6 +17,16 @@ static lsp::DocumentOnTypeFormattingResult processOnTypeFormatting(Fixture* fixt
     return fixture->workspace.onTypeFormatting(params);
 }
 
+TEST_CASE_FIXTURE(Fixture, "on_type_formatting_disabled_by_default")
+{
+    auto [source, marker] = sourceWithMarker(R"(
+        print("aaa {|")
+    )");
+
+    auto edits = processOnTypeFormatting(this, source, marker);
+    REQUIRE(!edits.has_value());
+}
+
 TEST_CASE_FIXTURE(Fixture, "on_type_formatting_converts_quotes")
 {
     client->globalConfig.format.convertQuotes = true;


lopi-py:
Mmm had to update the toml version, it was giving me errors at compile time 😓

<details>

```
JsonTomlSyntaxParser.cpp.o
In file included from /home/lopy/workspace/luau-lsp/extern/toml/include/toml11/toml.hpp:34,
                 from /home/lopy/workspace/luau-lsp/src/include/LSP/JsonTomlSyntaxParser.hpp:4,
                 from /home/lopy/workspace/luau-lsp/src/JsonTomlSyntaxParser.cpp:1:
/home/lopy/workspace/luau-lsp/extern/toml/include/toml11/toml/serializer.hpp: In member function ‘std::string toml::serializer<Value>::operator()(floating_type) const [with Value = toml::basic_value<toml::discard_comments, std::unordered_map, std::vector>]’:
/home/lopy/workspace/luau-lsp/extern/toml/include/toml11/toml/serializer.hpp:222:22: error: null destination pointer [-Werror=format-truncation=]
  222 |         std::snprintf(buf.data(), buf.size(), fmt, this->float_prec_, f);
      |         ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[3]: *** [CMakeFiles/Luau.LanguageServer.dir/build.make:275: CMakeFiles/Luau.LanguageServer.dir/src/JsonTomlSyntaxParser.cpp.o] Error 1
make[2]: *** [CMakeFiles/Makefile2:196: CMakeFiles/Luau.LanguageServer.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:241: CMakeFiles/Luau.LanguageServer.CLI.dir/rule] Error 2
make: *** [Makefile:137: Luau.LanguageServer.CLI] Error 2
```

</details>

JohnnyMorganz:
Looks reasonable to me. I think this behaviour should be gated behind a configuration option though (should it be off-by-default too?)

How value is the unfinished-string implementation? Seems like it would be a lot simpler if we skipped that. Atleast in vscode, normally the strings are complete when you start writing them already.

> Can this potentially replace the implementation of autocompleteEnd? it seems to relly too much on specific vscode behavior currently

I'd be open to changing this if it works well. The important part is that it should keep the cursor in the expected position, which is where the custom VSCode logic comes in

lopi-py:
> Looks reasonable to me. I think this behaviour should be gated behind a configuration option though (should it be off-by-default too?)

It *should* already be off by default in most editors, for example in VSCode you need to turn on `format.onType` (can't remember the exact name but it's something like that). I think it would make more sense to add an option to toggle this if we had more trigger characters

> How value is the unfinished-string implementation? Seems like it would be a lot simpler if we skipped that. Atleast in vscode, normally the strings are complete when you start writing them already.

I've simplified/refactored the whole logic so it should be easier to read/maintain. Also the user could have auto-pairs disabled or not have it as a built-in feature in its editor

> I'd be open to changing this if it works well. The important part is that it should keep the cursor in the expected position, which is where the custom VSCode logic comes in

We receive [formattingOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#formattingOptions) so we could try to determine the expected position maybe? Although this might be a breaking change and require users to enable `onTypeFormatting` in their editor

If we accept the breaking change we can also deprecate `completion.autocompleteEnd` and add something like `formatOnType.completeEnd/convertQuotes`

JohnnyMorganz:
> It should already be off by default in most editors, for example in VSCode you need to turn on format.onType (can't remember the exact name but it's something like that). I think it would make more sense to add an option to toggle this if we had more trigger characters

But a user might have `format.onType` enabled for other reasons (e.g., to *actually* perform formatting on their file, contributed by another extension, not to perform a potentially unexpected quote-change)

lopi-py:
Fair, just added the new option off by default

tests/OnTypeFormatting.test.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    lsp::DocumentOnTypeFormattingParams params;
+    params.textDocument.uri = uri;
+    params.position = {0, 12}; // after the '{'
+    params.ch = "{";
+    params.options = {4, true};
+
+    auto edits = workspace.onTypeFormatting(params);
+    REQUIRE(edits.has_value());
+    REQUIRE(edits->size() >= 1);
```
JohnnyMorganz:
```suggestion
    REQUIRE(edits->size() == 2);
```

Or are we expecting more than 2 edits? Same in the code down below

Also, not sure if it helps to visualise better, but could extract the `applyEdit` function out of Rename.test.cpp to test the final result.

JohnnyMorganz (commented):
Overall looks good. Can you revert the extra changes to toml library?

Also, please update the changelog

src/operations/OnTypeFormatting.cpp
```diff
@@ -0,0 +1,125 @@
...
+    auto lineText = textDocument->getLine(position.line);
+    size_t column = position.column;
+
+    auto doublePos = findQuote(lineText, column, '"');
+    auto singlePos = findQuote(lineText, column, '\'');
+    if (!doublePos && !singlePos)
+        return std::nullopt;
+
+    size_t quotePos = std::min(doublePos.value_or(SIZE_MAX), singlePos.value_or(SIZE_MAX));
+    size_t contentPos = quotePos + 1;
```
JohnnyMorganz: Not sure if it's worth considering this case, but this heuristic may fail if there are multiple strings on a single line (also if they are some or with mixed quotes), e.g. `x = "foo"; y = "bar {` or `x = "foo"; y = 'bar {` - we should look for the quote closest to the cursor.

lopi-py: `findQuote` goes from `column` to the left (right-to-left) so in cases like `x = "foo"; y = "bar {` it would find the first quote left to the caret. For the second case it would pick the `"` because it's picking the first quote (found 2 at col 9 and col 16), indeed we need to fix that. Another edge case I found is **x = `" {**

tests/OnTypeFormatting.test.cpp
lopi-py: > but could extract the applyEdit function out of Rename.test.cpp to test the final result.

Yes it helps to visualize better the result

lopi-py:
What if we have
```luau
x = "foo ` {|"
```
do we want to skip that (current behavior) or convert quotes to backtick and escape the existing ones?
```luau
x = `foo \` {|`
```

lopi-py:
Only edge case I can think is
```lua
foo "bar {|"
```
It would ignore the fact that its a function call and would produce invalid syntax. Do we want to handle this?

lopi-py:
> Mmm had to update the toml version, it was giving me errors at compile time 😓
> Details

Fixed this by compling with `clang++` instead of `gcc`

JohnnyMorganz (approved):
Thanks! Happy to merge the current setup and iterate on any further improvements

PR #1333, opened 2026-01-08
Cleanup InstanceFileSyncService handling

zovits:
No more need for the `:: any` typecasting. Additionally, uses `pcall` in a couple spots that were previously missed, which should fix the "InstanceFileSyncService is not enabled" issue and make this more robust overall.


JohnnyMorganz (approved):
Thanks!

PR #1335, opened 2026-01-12
Fix source transformation resolution, JSON deserialization, & transformed source position mapping

hello-42:
I ran into a few issues consecutively when attempting to use the plugins feature:

1. Any time a non-empty edits table was returned from a plugin's `transformSource` callback, the process for viewing the internal source would break. This told me that something relevant to transforming the source from the edits was breaking. The fix for this was to extract the transformed source prior to `pluginDoc` and then map the transformed source into the document. 

2. When using `lsp.json.deserialize()`, deserializing my `sourcemap.json` was causing an unexpected stack overflow for the table returned by deserialization. The fix to the error being emitted internally by `luau-lsp` below was to check if there's space on the stack:
- `lapi.cpp(813): ASSERTION FAILED: L->top < L->ci->top`

3. Post-transformation, `TextDocument`s and `PluginTextDocument`s would reference the transformed source's `Location` data to act as a source for the `completions`, `definitions`, `diagnostics`, `inlayHints`, and `signatureHelp` features - which led to the error listed below being emitted internally by `luau-lsp` for each feature, respectively. The fix was to return the original source's `Location` data instead.
- `Received response 'textDocument/[...] - (297)' in 3ms. Request failed: invalid string position (-32603)`

Some notes:
- I used AI (Opus 4.5) to make these changes. I don't know how to write C++, and I simply told my agent what issues I was facing. With that being said, I'm making this PR without the expectation that it'll be merged - mainly just to show what changes were made that helped me resolve these issues.
- I fed my agent any errors that I could witness coming from the binary's output, and lo and behold, these changes resulted and I was able to produce a build that works, supports the plugin functionality, and ultimately results in providing the intellisense I was looking for.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,6 +139,7 @@ target_sources(Luau.LanguageServer.Test PRIVATE
         tests/RequireGraph.test.cpp
         tests/OnTypeFormatting.test.cpp
         tests/AnalyzeCli.test.cpp
+        tests/Plugin.test.cpp
         tests/CodeAction.test.cpp
         tests/Plugin.test.cpp
         src/AnalyzeCli.cpp # For testing


JohnnyMorganz:
Thanks for the fixes here, I've folded them into #1088 and made a slight tweak to the last one by simplifying PluginTextDocument to be a composition

PR #1341, opened 2026-01-17
Process Roblox definitions before others in analyze CLI

Crystalflxme:
This matches the behavior of the rest of the language server. The implementation was taken from [here](https://github.com/JohnnyMorganz/luau-lsp/blob/ac76368826f0f5842ad9f063a388301b3e04dead/src/Workspace.cpp#L489-L498).


JohnnyMorganz:
Thanks, can you please update the changelog?

PR #1345, opened 2026-01-23
Fix editor guide

diamo-dev:
I have spent like an hour figuring out that its `--definitions` and not `--definition`. Yes, it's literally a single letter push but if I can save even one person an hour of searching through protocol specifications and forums, then this push has fulfilled its' purpose.


JohnnyMorganz (approved):
Thanks, and sorry for the confusion 🙁

PR #1362, opened 2026-02-07
fix: check `sourcemap.enabled` before autogenerating sourcemap server side

lopi-py:
Also mark external include directories as SYSTEM in CMake to suppress compiler warnings from third-party headers.

image


JohnnyMorganz:
Thanks! Mind adding a small fix line to the changelog?

lopi-py:
Done

Issue #1357, opened 2026-02-02
Show basic @luau API Documentation when platform is set to standard


EternalAbby:
hi

I'm not sure if this is a problem or this is just me. I'm using Linux and VSCode.

When choosing the platform type as `"standard"`, the API documentations are missing. From what I've seen, it might be because the `api-docs.json` file isn't configured into the `"luau-lsp.types.documentationFiles"` in the user settings.

To fix this, just manually add it into the settings.

if this isn't an actual issue then that's my bad lol

JohnnyMorganz:
If your platform type is `standard`, you are saying that you aren't using Roblox type definitions. `api-docs.json` contains mostly Roblox-specific type definitions, which is why it's not loaded.

If you're working with Roblox, you should set your platform to `roblox`.

In this case, I guess the main issue is that we lose the documentation for the built-in Luau functions?

EternalAbby:
yea, you're right

The main issue is that the documentation doesn't appear for the built-in Luau functions when the platform type is `standard`. My fix for it was to simply add in `api-docs.json` and it just worked as expected.

I'm not sure if there was supposed to be another file specifically for only built-in Luau functions (I haven't looked closely into this repo,) or if it should appear out-of-the-box, so that's all I could really say about this issue right now.

EternalAbby:
Hi @JohnnyMorganz.

I just looked through the commit to check out the new additions, very nice so far. One small issue, I saw that the "learn more" links are directing to the Roblox Creator Hub Documentation site (`create.roblox.com/docs`) instead of the Luau site (`luau.org`). This might not be ideal since the user had switched the platform to `"standard"` already.

Personally, it *might* be better to change the links to direct to the `luau.org` site instead.

Since this is generated from the roblox-specific `api-docs.json`, here are some quick changes that could be done:

- Replace `https://create.roblox.com/docs/reference/engine/libraries/` with `https://luau.org/library/`.
- There isn't an easy way to link to the library functions, only the library section, since these are block quotes instead of headers. In my opinion, it might be easier to remove `#.*` at the end of these links.
- Replace `https://create.roblox.com/docs/reference/engine/globals/LuaGlobals#` with `https://luau.org/library/#global-functions`.
- Again, there isn't an easy way to link to the functions, so it might be easier to remove `#.*`.

JohnnyMorganz:
[...]

PR #1386, opened 2026-02-28
Improvements for extracting @luau documentation

EternalAbby:
Since it's harder to correctly determine where the library functions are on the Luau site; they are put into block quotes, this pull request mainly adds changes to the URL's fragment so that it directs to the library's section.

- Changed links from directing to the roblox docs site to directing to the luau docs.
- Corrected `vector` library having `@roblox` prefix instead of `@luau` in the luau generated json.
- Import only the necessary methods.
- Pretty-print the JSON file.

This pull request is done to add ideas from #1357.


EternalAbby:
I'm not sure if I need to do the commits with the `luau-en-us.json` file generated or not.

also, ignore the first commit being unverified, forgot to add signing key

JohnnyMorganz:
Unfortunately I can't accept the changes to `scripts/api-docs/en-us.json` as that comes direct from upstream source. But we can look at the changes to the python generator and the generated `luau-en-us.json` file

EternalAbby:
first time using pull requests so this is quite jank, sorry about that

JohnnyMorganz (approved):
Thanks!

PR #1393, opened 2026-03-08
luau-lsp.sourcemap.includeNonScripts now works for non-rojo plugin environment

SkeletalReality:
The luau-lsp.sourcemap.includeNonScripts parameter used to only apply to the sourcemap generation in Rojo environments, not to the sourcemap generation by the Studio plugin. This provides that functionality


JohnnyMorganz:
Include non scripts is the default, so I fear this might blow up the sourcemap size for plugin synced data.

Also I don't think we need to bother changing "containsFilePaths" right? The check should only be at toJson level

PR #1443, opened 2026-04-11
Fix FindFirstAncestor when ancestor name is a substring of a child folder name

GaryJX:
## Problem

`getAncestorPath` in `src/Utils.cpp` uses a single `rfind(ancestorName + "/")` to locate the ancestor in the path string. When a folder in the path contains the ancestor name as a **substring** (e.g., a folder named `PrefixFoo` contains `Foo`), `rfind` matches the substring occurrence, the boundary check fails (character before is not `/`), and the function **returns `nullopt`** instead of continuing to search backward for valid occurrences.

This causes `FindFirstAncestor("Foo")` to fail for any script under folders whose name ends with `Foo` (like `PrefixFoo/`), producing "Unknown require: unsupported path" errors.

### Example

```
Path: game/Packages/_Workspace/Foo/Foo/Components/PrefixFoo/PrefixFoo
Searching for: "Foo/"
```

`rfind("Foo/")` matches inside `"PrefixFoo/"` (suffix match). The boundary check sees `x` (from "Prefix") instead of `/` and gives up. The valid match at `_Workspace/Foo/Foo/` is never found.

Packages without this substring collision work fine — e.g., `FindFirstAncestor("Bar")` resolves correctly when child folders are named `BarBaz` (since `"Bar/"` is not a substring of `"BarBaz/"`).

## Fix

Loop `rfind` backward until a boundary-valid match is found, instead of giving up on the first failed boundary check.

## Test Plan

- Added 3 test cases to `tests/Utils.test.cpp`:
  - Ancestor name as suffix of a child folder name (e.g., `PrefixFoo/` contains `Foo/`)
  - Ancestor name as prefix of a child folder name (e.g., `FooBar/` — verifying `Foo/` does NOT falsely match)
  - Ancestor name as suffix with a different prefix (e.g., `FooBar/` contains `Bar/`)
- All existing `getAncestorPath` tests continue to pass

## Verified

Reproduced via CLI — a file using `FindFirstAncestor` under a folder with a substring collision reports `TypeError: Unknown require: unsupported path`. The same file with `script.Parent.Parent...` resolves correctly (confirming the sourcemap is valid, only `FindFirstAncestor` resolution is broken).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed `FindFirstAncestor` failing when ancestor name is a substring of another folder name in the path (e.g., `FindFirstAncestor("Foo")` failed when a folder named `PrefixFoo` existed in the path)
 - When a plugin is hot-reloaded, all source nodes (not just managed text documents) are now marked dirty so non-managed files are re-analysed with the updated plugin transformations
 - Plugin transformations are no longer applied to plugin files themselves ([#1433](https://github.com/JohnnyMorganz/luau-lsp/issues/1433))
 - Fixed string-require auto imports in Roblox mode ignoring user-defined aliases from `.luaurc`. Aliases are now preferred over `@game/...` paths when available ([#1436](https://github.com/JohnnyMorganz/luau-lsp/issues/1436))
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -34,15 +34,21 @@ std::optional<std::string> getAncestorPath(const std::string& path, const std::s
     // Append a "/" to the end of the parentPath to make searching easier
     auto parentPathWithSlash = *parentPath + "/";
 
-    auto ancestor = parentPathWithSlash.rfind(ancestorName + "/");
-    if (ancestor != std::string::npos)
+    auto searchTarget = ancestorName + "/";
+    auto searchFrom = parentPathWithSlash.size();
+    while (searchFrom > 0)
     {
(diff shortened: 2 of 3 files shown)


GaryJX:
Hello! Disclaimer, this PR was generated using AI (not sure if that's welcome or not, and if this is already a known issue. If not welcome, feel free to close). I am a user of the extension (thank you for building this, it's been really helpful!) and found a niche bug where the LSP would fail to properly path find using FindFirstAncestor when a parent folder name / file name ends with the ancestor name (e.g. folder/file name is `FooBar`, and ancestor name is `Bar`)

JohnnyMorganz:
Hey! Thanks for the fix.

Just wondering - would it be simpler to just keep the original logic and search for `/ancestorName/` instead of `ancestorName/`?

There's an edge case where the ancestor we're seatching for is the start of path (game/), but I think that's already broken (is the ancestor actually named game in the place file?). If we really want to fix that, we could add a starts with check to handle it.

Using AI is not a problem, and is welcome, provided the change is well thought out. No need to take away a useful tool

GaryJX:
> Hey! Thanks for the fix.
> 
> Just wondering - would it be simpler to just keep the original logic and search for `/ancestorName/` instead of `ancestorName/`?
> 
> There's an edge case where the ancestor we're seatching for is the start of path (game/), but I think that's already broken (is the ancestor actually named game in the place file?). If we really want to fix that, we could add a starts with check to handle it.
> 
> Using AI is not a problem, and is welcome, provided the change is well thought out. No need to take away a useful tool

Makes sense! Updated the PR to reflect the changes

JohnnyMorganz (approved):
Thank you!

PR #1463, opened 2026-04-22
Implement chained alias support

vrn-sn:
Implements this amendment to the require-by-string aliases RFC: https://github.com/luau-lang/rfcs/pull/145. Aliases in a given configuration can reference each other, and cycles are handled gracefully with a `std::nullopt` result.

I noticed that there's an existing TODO for reporting alias resolution errors, so I left an additional TODO for reporting alias cycle errors. I'm assuming whatever reporting mechanism is used to solve the first problem can easily be used for the second problem as well.


vrn-sn:
Oops totally missed that you did this a few days ago @JohnnyMorganz. Closing mine!

Feel free to take anything you want from this (e.g. test cases) and bring them into #1460 if you find it useful.

PR #1472, opened 2026-04-30
Repair broken Tuple/typepack support.

MaximumADHD:
This commit sanitizes an issue on Roblox's end where it's possible for illegal typepacks to be made for backend events (maybe also functions and callbacks too, but less likely).

It applies the following workarounds:
* Multiple tuples declared in a parameter list are no longer valid, all but the last one will now get normalized into `any`.
* If the singular `Tuple` isn't the last parameter, it is considered invalid and gets normalized to `any`.


JohnnyMorganz (approved):
Seems reasonable to me

PR #1480, opened 2026-05-04
Fix inaccurate `require()` description when platform is "standard."

### LINKED ISSUE #1479: `require()` description is inaccurate when platform is "standard."
When switching the platform to "standard," `require()` preserves the description from its Roblox Studio counterpart. This doesn't make sense from an outside perspective as the information are written for that platform only.

Because there aren't any official description of `require()` from base Luau, since it relies on runtimes to implement this function and Lua 5.1's `require()` is different from Luau's, it might be a good idea to write an unofficial description for this function.

As a reference point, just in case, base Luau's `require()` uses require by string semantic via relative paths and aliases defined in `.luaurc` or `.config.luau`.

EternalAbby:
This pull request adds an unofficial description for `require()` when platform is "standard" and regenerated the JSON.

This description is based off of the original official description.

It also has a minor refactor over the script, which shouldn't change anything else.

Resolves #1479.


EternalAbby:
uh idk how to remove the previous commits, hope that it's fine

scripts/extractLuauDocs.py
```diff
@@ -1,67 +1,69 @@
...
-    # Otherwise, it's from a global function.
+    # otherwise, it's from a global function. cannot do more than this.
     else:
         luau_docs[docs_key]["learn_more_link"] = "https://luau.org/library/#global-functions"
 
+# replacing `require()` description with a general one. this new description is based off of the original description.
+luau_docs["@luau/global/require/param/0"]["documentation"] = "The directory path to a file or a folder/directory to retrieve the return value it provides. Directory path can use aliases defined in <code>.luaurc</code> or <code>.config.luau</code>. If the path is a file, executes that file. If the path is a folder/directory, finds <code>init.luau</code> or <code>init.lua</code> and execute that instead."
+luau_docs["@luau/global/require/return/0"]["documentation"] = "What the file or folder/directory returned (usually a table or a function)."
+luau_docs["@luau/global/require"]["documentation"] = "Returns the value that was returned by the given directory path, running it if it has not been run yet."
+
```
JohnnyMorganz, line 67:
```suggestion
# replacing `require()` description with a general one. this new description is based off of the original description.
luau_docs["@luau/global/require/param/0"]["documentation"] = "The path to a file or a directory that will be executed to retrieve the return value it provides. Paths may use aliases defined in <code>.config.luau</code> or <code>.luaurc</code>. If the path is a file, executes that file. If the path is a directory, finds an <code>init.luau</code> or <code>init.lua</code> file inside of that directory to execute."
luau_docs["@luau/global/require/return/0"]["documentation"] = "What the file returned (usually a table or a function)."
luau_docs["@luau/global/require"]["documentation"] = "Returns the value that was returned by the given path, running it if it has not been run yet."

```

JohnnyMorganz (commented):
Minor rewording suggestion

scripts/extractLuauDocs.py
EternalAbby: Thanks! I had trouble coming up with a good description that includes everything about `require()`, so it ended up having a lot of unnecessary words.

JohnnyMorganz (approved):
Thanks

PR #1482, opened 2026-05-07
feat: implement server/client boundary filtering for auto-imports

### LINKED ISSUE #1065: Auto complete permitted in environment that forbids specific module
Client can autocomplete modules in ServerStorage

LDGerrits:
Added a new optional setting, `luau-lsp.completion.imports.serverClientFiltering.enabled`, to enforce the Roblox server-client boundary for auto-import suggestions.

When enabled, it will analyze the context of the script being edited and the location of the module being suggested. If a suggestion crosses a Roblox boundary (such as a client script attempting to import a server-only module), the suggestion is filtered out to reduce noise and prevent invalid code.

## Container Categorization
To enforce this boundary, I categorized the containers as follows:

**Server-Only Containers:**
- ServerScriptService
- ServerStorage
- Scripts with the .server.lua or .server.luau extension.

**Client-Only Containers:**
- StarterPlayer (including StarterPlayerScripts and StarterCharacterScripts)
- StarterGui
- StarterPack
- ReplicatedFirst
- Scripts with the .client.lua or .client.luau extension.

**Shared Containers:**
- `ReplicatedStorage`, `workspace`, and all other top-level containers are treated as Shared. Modules here can be suggested to both the client and the server.

## Tests
Added 8 unit tests to AutoImports.test.cpp covering:
- Client-to-Server filtering (should hide)
- Server-to-Client filtering (should hide)
- Shared-to-Any visibility (should show)
- Toggle correctly enables/disables the feature

```bash
[doctest] test cases:   810 |   810 passed | 0 failed | 0 skipped
[doctest] assertions: 65456 | 65456 passed | 0 failed |
[doctest] Status: SUCCESS!
```


JohnnyMorganz (commented):
Thanks for the contribution! I like the idea, turns out to be pretty simple in practice. A couple thoughts:

- Can you think of any reason why someone would *not* want this enabled by default? I don't think we need to bother with an optional setting
- Should `getScriptContext` just be a function on `SourceNode`? Seems like it makes sense there. And we can have some unit tests just testing the output of getScriptContext in the different cases. Will also prevent the duplication between the two auto importers.

JohnnyMorganz:
This will close https://github.com/JohnnyMorganz/luau-lsp/issues/1065

LDGerrits:
The optional setting was probably playing it too safe. I can only imagine rare cases where developers put a `ModuleScript` inside of `StarterGui` or `ReplicatedFirst` and want to require them on the server (technically, the server is able to use these containers). However, I really can't imagine a reason why a developer would be structuring their code like this. I think this is a very sane default that reduces noise for 99.9% of developers.

As per your suggestion, I have moved `getScriptContext()` to `RobloxSourceNode.cpp`, and managed to simplify and optimize some of the code. Also, I added more test cases to `AutoImports.test.cpp` (14 in total), and added 1 test specifically for `getScriptContext()` in `Sourcemap.test.cpp`.

[doctest] test cases:   817 |   817 passed | 0 failed | 0 skipped
[doctest] assertions: 65502 | 65502 passed | 0 failed |
[doctest] Status: SUCCESS!

JohnnyMorganz (approved):
Thanks! I made a couple tweaks to simplify the logic a bit, moving the string-require filtering to RobloxFileResolver and pre-computing script context

PR #1512, opened 2026-05-21
Fix @self path alias resolution for sourcemap virtual nodes

### LINKED ISSUE #1511: LSP incorrectly resolves @self to the inner $path directory rather than the project root
When working on a plugin or library project, it fails to resolve dependencies mapped alongside the main source directory when using the `@self/` path alias.

It appears the LSP evaluates `@self` as starting from the `$path` directory of the root node (in this case, `src/`), rather than the actual project root that contains the mapped sibling directories. As a result, requiring a sibling package via `@self/packages/test` works perfectly at runtime, but fails to resolve in the LSP.

**Minimal Project Structure:**

```text
default.project.json
packages/
  test.luau
src/
  init.server.luau
sourcemap.json
```

**Steps to Reproduce:**

1. Create a `default.project.json` where a folder (`packages`) is at the same level as `src`:

```json
{
  "name": "test",
  "tree": {
    "$path": "src",
    "packages": {
      "$path": "packages"
    }
  }
}
```

2. Generate the sourcemap using Rojo or Argon. The resulting `sourcemap.json` looks like this:
```json
{
  "name": "test",
  "className": "Script",
  "filePaths": ["src/init.server.luau", "default.project.json"],
  "children": [
    {
      "name": "packages",
      "className": "Folder",
      "children": [
        { "name": "test", "className": "ModuleScript", "filePaths": ["packages/test.luau"] }
      ]
    }
  ]
}
```

3. Inside `src/init.server.luau`, attempt to require the package:

```luau
-- Works at runtime, but LSP fails to resolve (thinks @self is `src/`)
local test = require("@self/packages/test")
```

4. To satisfy the LSP, you have to use a workaround like `@game`, which then breaks the actual runtime:

```luau
-- Resolves in LSP, fails at runtime
local test = require("@game/packages/test")
```

**Expected Behavior:**
`luau-lsp` should correctly resolve the `@self/packages/test` path based on the sourcemap hierarchy, recognizing that `@self` refers to the root `test` node (which contains both `src` and `packages`), not just the `src` folder.

**Actual Behavior:**
The LSP assumes `@self` starts from inside `src`. Because of this, it cannot see the `packages` directory. This also breaks relative paths, even if a script is not an `init` file, doing something like `require("./packages/...")` fails to traverse correctly.

**Environment:**
* **OS:** Arch Linux
* **Editor:** Helix (also reproduced on VSCode)
* **luau-lsp version:** 1.68.0
* **Sourcemap generator:** Argon 2.0.29 / Rojo 7.6.1

0hirume:
Fix `@self` alias resolution for non-DataModel sourcemap roots

When a project builds an `rbxm` (e.g. a plugin or library) rather than a full `DataModel`, the sourcemap root is the script itself rather than a `DataModel` node. In this case, sibling directories mapped alongside the `$path` directory are children of the root sourcemap node, not children of a filesystem directory.

Previously, `@self` on a virtual path context was delegated to `LSPPlatform::resolveStringRequire`, which resolves `@self` using the filesystem path of the calling file as the base URI. For an `init.server.luau` inside `src/`, this meant `@self` resolved to `src/`, making sibling directories like `packages/` invisible to the LSP, even though they're valid children of the root sourcemap node at runtime.

The fix handles `@self` directly in `RobloxPlatform::resolveStringRequire` when the context is already a virtual path, using the sourcemap node's virtual path as the base instead of the filesystem path. User-defined aliases are unaffected and still delegate to `LSPPlatform` as before. Non-sourcemap files are also unaffected since they never reach this code path.

Fixes #1511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed `@self` string-require aliases resolving from the filesystem instead of the sourcemap tree for non-DataModel roots ([#1511](https://github.com/JohnnyMorganz/luau-lsp/issues/1511))
+
 ## [1.69.0] - 2026-07-14
 
 ### Added
--- a/src/platform/roblox/RobloxFileResolver.cpp
+++ b/src/platform/roblox/RobloxFileResolver.cpp
@@ -127,7 +127,15 @@ std::optional<Luau::ModuleInfo> RobloxPlatform::resolveStringRequire(
         std::string aliasName = requiredString.substr(1, slashPos == std::string::npos ? std::string::npos : slashPos - 1);
         std::string aliasNameLower = toLower(aliasName);
 
-        if (luauConfig.aliases.find(aliasNameLower) || aliasNameLower == "self")
+        if (aliasNameLower == "self")
+        {
+            std::string remainder = (slashPos == std::string::npos) ? "" : requiredString.substr(slashPos + 1);
+            if (remainder.empty())
+                return Luau::ModuleInfo{context->name};
+            return Luau::ModuleInfo{context->name + "/" + remainder};
+        }
+
+        if (luauConfig.aliases.find(aliasNameLower))
             return LSPPlatform::resolveStringRequire(context, requiredString, limits);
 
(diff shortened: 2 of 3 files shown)


JohnnyMorganz (approved):
Thanks! Could you update the changelog?

0hirume:
Updated, thanks!

JohnnyMorganz (approved):

PR #1517, opened 2026-05-27
fix Can't convert to JSON

IAM9PASA:


JohnnyMorganz:
I think this was resolved by https://github.com/JohnnyMorganz/luau-lsp/pull/1559

PR #1532, opened 2026-06-08
.config.luau typing

nnullcolumn:
full disclosure, this PR was vibecoded with some human review / revisions, some code may be unsatisfying. this is pretty self-contained implementation wise so i'm not very worried about it but if any changes are needed let me know.
- autocompletion works for .config.luau
- type-checking works for .config.luau
- added an environment for .config.luau which exports a few useful types:
  - LanguageMode
  - LintWarning (note, this does not include "*" despite it being allowed in the table. This can be trivially changed but I figure it is easier for users to add to an alias than remove from one which actually wants something specific)
  - Config
  - LuauConfig (would we prefer people do `index<Config, "luau">` instead?)

things i, a human, verified, corrected, or added test cases for:
- these features actually work in my codebase when i build to vsix and reasonable tests are included
- test case namings follow ~~conventions~~ surrounding code ❤️‍🩹
- fixed some error messages and logs to be clear
- random renames in the codebase that did not need to happen (woah, AI agents are doing this now? I know they're prone to hallucinating random BS solutions, but this one was a style choice! very interesting.)
- .config.luau environment does not leak types or return rules into regular files
- .config.luau types work for any valid config and are checked
- .config.luau applies to the .config.luau docu-ment itself (i imagine this is desired?)
- it does not apply to files named `foo.config.luau`, only files named `.config.luau` (is this desired/specified behavior?)

extra notes:
- environment globals and definitions are the same as the regular document env. i left it this way because `typeof(require(...))` is valid & useful, and because `.config.luau` may get forms of import support in the future, but if stricter safeguards are needed let me know and i'll fix it


src/ConfigLuau.cpp
```diff
@@ -0,0 +1,168 @@
...
+        ss << "}\n";
+        ss << "declare function ";
+        ss << kCheckFunctionName << "(config: Config): Config";
+        ss << "\n";
+
+        return ss.str();
+    }();
+
+    return definitions;
+}
```
nnullcolumn, line 59: human here! this ai code kinda sucks, but seems like the move for this declaration in particular since it's so programmatic. if there is a better way to auto-generate and pull this declaration i am all ears?

PR #1555, opened 2026-06-30
Add more Moonwave documentation support

### LINKED ISSUE #1552: Function hover hints should display certain moonwave tags in a more semantic manner
Moonwave docs support various tags. It may be worthwhile to build out some of these as function hover hints in a more semantic manner.

Some of the ones considered that are really useful are:
- `@yields`
- `@private` - this particularly should signal to the user that this particular method is private. for the sake of cross-compatibility with moonwave, I don't think it's wise to build a `@protected` tag unless you can bridge that gap, nonetheless it'd be very cool.
- `@deprecated`
- `@readonly`

This ticket should be labeled as feature/improvement.

ryancundiff:
Add more Moonwave documentation support to the tooltips. This relates to Issue #1552.

**Readable flag tags.** Previously, `@yields` and `@unreleased` were the only tags parsed for bolding, and they rendered as the raw tag name (e.g. **@yields**). Flag tags now render as readable labels instead, and I added parsing for several more. Like `@yields` and `@unreleased`, these inform the user of the intended usage context of the associated item:

- `@server` → **Server**
- `@client` → **Client**
- `@plugin` → **Plugin**
- `@private` → **Private**
- `@readonly` → **Read Only**

**`@deprecated` support.** The tag is parsed into its version and description rather than printed as plain comment text:

```
--[=[
  @function oldAdd
  Adds two numbers together.

  @param a number
  @param b number -- Number to be added to `a`.

  @return number -- Sum of `a` and `b`.

  @since v1.0.0
  @deprecated v1.2.0 -- Use `newAdd` instead.
]=]
```

The body of the tooltip now renders as:

> Adds two numbers together.
>
> **Parameters**
> - `a` number
> - `b` number -- Number to be added to `a`.
>
> **Returns**
> - `number` -- Sum of `a` and `b`.
>
> **Since** `v1.0.0`
> **Deprecated** `v1.2.0` -- Use `newAdd` instead.

**`@since` support.** As shown above, `@since` renders as a bolded **Since** notice with the version in an inline code span.

**`@field` support.** `@field` entries are collected into a **Fields** list, mirroring the existing `@param` handling — useful for `@interface` doc comments.

**Ignored declaration and metadata tags.** Tags that only carry meaning for the documentation site generator (`@class`, `@function`, `@method`, `@prop`, `@interface`, `@type`, `@__index`, `@external`, `@tag`, `@within`, `@ignore`) are no longer printed as raw text in the tooltip.

Closes #1552.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Sync to upstream Luau 0.726
+- Sync to upstream Luau 0.728
 - Moonwave flag tags (e.g. `@yields`, `@server`, `@private`) are now rendered as readable labels (e.g. **Yields**) instead of the raw tag name
 - Moonwave declaration and metadata tags (`@class`, `@function`, `@method`, `@prop`, `@interface`, `@type`, `@__index`, `@external`, `@ignore`) are no longer shown as raw text in hover documentation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Added support for more Moonwave tags in hover documentation: `@deprecated` and `@since` are now rendered with their version and description, `@field` entries are shown in a "Fields" list, and `@readonly` is boldified
+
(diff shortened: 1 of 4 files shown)


JohnnyMorganz (approved):
Thank you!

ryancundiff:
I ran the tests locally, and they all pass for me now.

jasonconba:
you guys rock, thanks!! @ryancundiff @JohnnyMorganz

PR #1558, opened 2026-07-03
Add configuration option to ignore `.lua` files

Ransomwave:
This PR addresses issue #1429.

Use-case: Standalone Lua development outside of Roblox. Although older Rojo setups generate `.lua` files that require luau-lsp processing, this is legacy behavior. Because the `.luau` extension is now the standard for Luau, its tooling should hopefully remain separate from standalone Lua environments.

* Added `"luau-lsp.analyzeLuaFiles"` boolean setting to enable/disable analysis of `.lua` files in addition to `.luau` files. The setting includes documentation and a default value of `true`.
* Updated `extension.ts` to respect the new `analyzeLuaFiles` setting, so the language server only attaches to `.lua` files if enabled.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -231,6 +231,12 @@
           ],
           "scope": "resource"
         },
+        "luau-lsp.analyzeLuaFiles": {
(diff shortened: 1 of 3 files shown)


JohnnyMorganz:
Thanks, can you add an entry in the changelog for this?

Ransomwave:
@JohnnyMorganz Added a changelog entry. Let me know if there's anything else I need to do.

JohnnyMorganz (approved):
Thanks

PR #1559, opened 2026-07-03
Add UTF-8 validation for instance names

MarsSquirrel:
# Problem
While developing with Luau-lsp, I ran into a confusing error where Luau-lsp could not convert the tree table into valid JSON, and threw a vague `Can't convert to JSON` error.

The root cause of this problem was incredibly obscure. One of the instances that Luau-lsp was trying to parse had a decorative Unicode character in its name that had been truncated mid-character by some sort of external tool, leaving an incomplete, corrupted UTF-8 byte sequence at the very end of the instance's name. Because Luau-lsp relies on JSON-RPC to communicate with Roblox Studio, this malformed trailing byte caused the entire serialization process to crash with zero context as to which instance was actually breaking it.

# Solution
To save anyone else who has this issue from hours of debugging a confusing JSON error, this PR introduces a clear error message to help developers quickly resolve this issue:
* Added instance name truncation detection - Created the function, `InstanceTracker.hasTrailingCorruptedBytes()` to check all instance names for incomplete byte sequences.
* Added a condition inside `InstanceTracker.encode()` to verify the instance name does not have corrupted bytes. If it does, an error log will be sent to the developer's console including the exact instance that is failing.

### DIFF
--- a/plugin/src/InstanceTracker.luau
+++ b/plugin/src/InstanceTracker.luau
@@ -170,10 +170,6 @@ function InstanceTracker.encode(
 ): types.EncodedInstance
 	Log.debug(`Encoding {instance:GetFullName()}`)
 
-	if not isValidUTF8(instance.Name) then
-		Log.warn(`Cannot encode instance "{instance:GetFullName()}" because its name contains invalid UTF-8.`)
-	end
-
 	local encoded: types.EncodedInstance = {
 		Name = instance.Name,
 		ClassName = instance.ClassName,
@@ -183,6 +179,11 @@ function InstanceTracker.encode(
 
 	for _, child in instance:GetChildren() do
 		if not self:shouldInclude(child, includeList) then
+			continue
+		end
+
+		if not isValidUTF8(instance.Name) then
+			Log.warn(`Cannot encode instance "{instance:GetFullName()}" because its name contains invalid UTF-8.`)
 			continue
 		end
 
--- a/plugin/src/InstanceTracker.luau
+++ b/plugin/src/InstanceTracker.luau
@@ -148,6 +148,29 @@ function InstanceTracker.shouldInclude(self: InstanceTracker, instance: Instance
 	return false
 end
 
+--- Checks if the string ends with a corrupted, malformed, or truncated UTF-8
+--- replacement character sequence (bytes 239, 191, 189)
+--- @param self -- The InstanceTracker instance.
+--- @param name -- The name of the instance.
+--- @return boolean -- True if the instance name contains trailing corruption bytes.
+function InstanceTracker.hasTrailingCorruptedBytes(self: InstanceTracker, name: string): boolean
+	local replacementChar = string.char(239, 191, 189)
+
+	if string.find(name, replacementChar .. "$") then
+		return true
+	end
+
+	if string.find(name, string.char(239) .. "$") then
+		return true
+	end
+
+	if string.find(name, string.char(239, 191) .. "$") then
+		return true
+	end
+
+	return false
+end
+
 --- Recursively encodes an Instance and its children into a serializable table.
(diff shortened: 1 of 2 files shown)


JohnnyMorganz:
How does this fix compare to https://github.com/JohnnyMorganz/luau-lsp/pull/1517?

The other PR seems to actually handle the utf8 character better rather than failing the sync completely. But I haven't had the chance to test it out myself yet

MarsSquirrel:
The other PR definitely handles sanitization better for all UTF8 issues, but I'm concerned with it for two reasons:
1. Their solution is just to sanitize the paths sent to the LSP, but wouldn't that make the paths not correct? The sanitized filenames wouldn't match the instances in studio. I'm not sure if this is exactly the case as I haven't tested their solution.
2. I think the LSP should warn users to fix this issue on sync, because Roblox already can't reference these broken instances by name without erroring, so if you sanitized the file paths you're really just kicking the can down the road.

I think that my PR is definitely one dimensional and only solves my specific issue (UTF8 byte sequences cut off), while their function certainly fixes more issues with UTF8. I'll look into this though.

MarsSquirrel:
#1517 handles the error by sanitizing file names and replacing bad UTF8 sequences with question marks, as shown here:
Screenshot 2026-07-07 at 1 32 02 AM
(The bad instance)

Screenshot 2026-07-07 at 1 32 47 AM
(Referencing the instance through Luau-LSP, the replacement character has been filled in with a question mark)

This honestly isn't a bad solution, but the worry for me is that Luau-LSP is serving incorrect filenames for autocompletion. Obviously, this would be immediately noticed at runtime:
Screenshot 2026-07-07 at 1 34 46 AM
But my concern is that Luau-LSP is now serving filenames that do not exist within the datamodel. While the error at runtime is pretty obvious once the code executes, the editor experience becomes misleading because Luau-LSP is referencing instances that do not exist.

I believe the LSP should serve the real filenames, and if it can't (because of issues like this), it should clearly inform the developer of the issue and why the LSP can't sync. 

Screenshot 2026-07-07 at 1 42 08 AM
<br>
You can't even reference these instances in the command bar or a script; Roblox just doesn't autocomplete them. So Luau-LSP should do the same, and instead of replacing the bad characters, it should stop the sync and clearly inform the developer of the issue.

I'd love to hear your thoughts and opinions on this :)

MarsSquirrel:
Simplified the string validation to handle all UTF8 errors, and not just mine.
Screenshot 2026-07-07 at 2 07 24 AM

plugin/src/InstanceTracker.luau
```diff
@@ -163,6 +170,10 @@ function InstanceTracker.encode(
 ): types.EncodedInstance
 	Log.debug(`Encoding {instance:GetFullName()}`)
 
+	if not isValidUTF8(instance.Name) then
+		Log.warn(`Cannot encode instance "{instance:GetFullName()}" because its name contains invalid UTF-8.`)
```
JohnnyMorganz: Should we skip the instance then so that it doesn't break the JSON encoding later?

Otherwise your approach seems reasonable to me

JohnnyMorganz:
Could you also update the changelog with this fix?

plugin/src/InstanceTracker.luau
MarsSquirrel: That's smart. However, I'm not entirely sure how we would skip it. We could move the condition to be inside the `GetChildren` loop, and just have it continue if the child's name is not valid utf8. This would mean that `game` would still be susceptible to this issue, but I doubt anyone is going to have this happen ¯\_(ツ)_/¯

I'll go ahead and implement it like this and update the changelog.

MarsSquirrel:
@JohnnyMorganz, I've updated the changelog. Please let me know if you need me to change anything else.

Issue #1591, opened 2026-08-18
Add shortest-path and alternative-path options for string-require auto-imports


kyrtapdev:
### Motivation
String-require auto-imports currently support these `luau-lsp.completion.imports.requireStyle` values:
- `auto`
- `alwaysRelative`
- `alwaysAbsolute`

When `"luau-lsp.platform.type": "roblox"`, `auto` uses DataModel hierarchy heuristics rather than comparing the generated strings. In particular, it generally chooses a relative path for targets in the same service subtree, even when a configured `.luaurc` alias would produce a shorter and more readable require.

For example, both imports may be valid:
```json
{
	"aliases": {
		"pkg": "src/Packages"	
	}
}
```
```luau
local Test = require("./../../../packages/Test")
local Test = require("@pkg/Test")
```
But `auto` may choose the relative form solely because both modules are in the same service subtree.

### Requested: `shortest` require style
Please add a `shortest` value for `luau-lsp.completion.imports.requireStyle`.
```json
{
	"luau-lsp.completion.imports.requireStyle": "shortest"
}
```
For example:
```luau
-- Relative candidate:
require("./../../../packages/Test")
-- Alias candidate:
require("@pkg/Test")
-- shortest result:
require("@pkg/Test")
```
This should apply equally to standard and Roblox string-require auto-imports where aliases are available.

### Requested: show valid path alternatives in completion
Please also consider a separate setting that exposes valid import-path alternatives as separate completion items:
```json
{
	"luau-lsp.completion.imports.showRequirePathAlternatives": true
}
```
For a module with both relative and alias paths available, completion could offer both:
```luau
local Test = require("./Test")
local Test = require("@pkg/Test")
```
This would let users select the intended import style per use case without changing the existing default behavior.

LordOfSpelunky:
Related: https://github.com/JohnnyMorganz/luau-lsp/issues/625

LordOfSpelunky:
https://github.com/JohnnyMorganz/luau-lsp/blob/main/src/platform/StringRequireAutoImporter.cpp#L25

Seems like it's supposed to do this at least

PR #1592, opened 2026-08-18
Add nearestAbsolute auto-import require style

### LINKED ISSUE #625: Make auto require more data model aware?
I'm experimenting with using Luau LSP internally at Roblox. Auto requires are fantastic, but we have a common pattern of the following:

```lua
local Plugin = script:FindFirstAncestor("PluginName")

local Script = require(Plugin.Src.Script)
```

Unfortunately, it looks like Luau LSP will auto-require without the Plugin part, and will just script.Parent etc, which we don't want.

I think that Luau LSP should (optionally?) use variables if they exist and statically point to the strict data model instance. That is, if we have some require already such that Luau LSP already knows where to put stuff, and it sees that the equivalent to script.Parent.Parent exists in the form of Plugin, it should use that.

I think another problem with trying to use auto requires exclusively for us is that I'm not sure Luau LSP will actually split itself off from the Plugin variable, but instead might try to put itself at the beginning. Would be cool to also be able to tell it to place the first require wherever it can find a strict data model instance?

LordOfSpelunky:
Wasn't sure if this should be part of absolute/relative, so decided instead to make it its own option. May be a bit worrisome that it is only used in Roblox, I thought maybe the same could be used for #1591? 

---

Requires modules relative to the nearest fixed variable that resolves to a known ancestor instance (e.g. `local Main = script:FindFirstAncestor("X")` or `local Main = script.Parent.Parent`), producing requires like `require(Main.X.Y)`. Anchor expressions are resolved by reusing the platform's magic module resolvers, and behaviour falls back to `auto` when no anchor is an ancestor of the module.

Closes #625.

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -624,12 +624,14 @@
           "enum": [
             "auto",
             "alwaysRelative",
-            "alwaysAbsolute"
+            "alwaysAbsolute",
+            "nearestAbsolute"
           ],
           "enumDescriptions": [
             "Automatically compute the style of require to use based on heuristics",
             "Always require the module relative to the current file",
-            "Always require the module absolute based on root"
+            "Always require the module absolute based on root",
+            "Require the module relative to the nearest fixed variable that resolves to a known ancestor instance (e.g. `local Main = script:FindFirstAncestor(\"PluginName\")` or `local Main = script.Parent.Parent`), falling back to `auto` when no such variable exists"
           ],
           "scope": "resource"
         },
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -631,7 +631,7 @@
             "Automatically compute the style of require to use based on heuristics",
             "Always require the module relative to the current file",
             "Always require the module absolute based on root",
-            "Require the module relative to the nearest fixed variable that resolves to a known ancestor instance (e.g. `local Main = script:FindFirstAncestor(\"PluginName\")` or `local Main = script.Parent.Parent`), falling back to `auto` when no such variable exists"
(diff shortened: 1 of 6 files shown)


editors/code/package.json
```diff
@@ -624,12 +624,14 @@
...
-            "alwaysAbsolute"
+            "alwaysAbsolute",
+            "nearestAbsolute"
           ],
           "enumDescriptions": [
             "Automatically compute the style of require to use based on heuristics",
             "Always require the module relative to the current file",
-            "Always require the module absolute based on root"
+            "Always require the module absolute based on root",
+            "Require the module relative to the nearest fixed variable that resolves to a known ancestor instance"
```
JohnnyMorganz:
```suggestion
            "Require the module relative to the nearest fixed variable that resolves to a known ancestor instance. Only applicable in Roblox projects"
```

src/platform/roblox/InstanceRequireAutoImporter.cpp
```diff
@@ -30,11 +32,64 @@ std::string optimiseAbsoluteRequire(const std::string& path)
     return path;
 }
 
+static std::optional<Luau::ModuleName> resolveFixedVariablePath(
```
JohnnyMorganz, line 35: I was hoping we might be able to just reuse `RequireTraceResult`'s `expr` store, but I don't think that would work here as the stored exprs are only populated for AST nodes reachable from an existing require, whilst this needs arbitrary variable mappings.

Seems reasonable to me

JohnnyMorganz (approved):
Thanks!

Issue #1594, opened 2026-08-19
Auto-import boundary: RunContext client scripts are classified as server


dev-bap:
Hit this after switching a project to `emitLegacyScripts: false`. Auto-import started offering me `ServerScriptService` modules from inside client scripts.

The context comes from className alone:

https://github.com/JohnnyMorganz/luau-lsp/blob/main/src/platform/roblox/RobloxSourcemap.cpp#L489-L494

```cpp
if (node->className == "Script")
    node->scriptContext = ScriptContext::Server;
else if (node->className == "LocalScript")
    node->scriptContext = ScriptContext::Client;
```

Once you're on RunContext there's no LocalScript any more. A `.client.luau` comes out as a plain `Script`, so it reads as server, and nothing in the sourcemap says otherwise. What gets suggested are requires that error at runtime.

It isn't specific to `emitLegacyScripts`. A sync rule using `runContextClientScript` produces the same `Script`, and so does a `.meta.json` setting `RunContext = Client` on a `.server.luau`. That last one is the annoying one: you've written the run context out in full and it still comes back as server.

Moving the script into StarterPlayerScripts doesn't help either. The className check runs before the ancestor name check, so a `Script` is Server wherever you put it.

Repro:

```json
{
  "name": "example",
  "emitLegacyScripts": false,
  "tree": {
    "$className": "DataModel",
    "ReplicatedStorage": { "Client": { "$path": "src/client" } },
    "ServerScriptService": { "Server": { "$path": "src/server" } }
  }
}
```

Put a half-typed identifier in `src/client/Main.client.luau` and a module under `src/server/`. `rojo sourcemap` gives `"className": "Script"` for Main, and completion inside it offers the server module.

As for fixing it, matching `.client.luau`/`.server.luau` against `filePaths` would handle the common case, which is close to what #1482 originally described. It won't catch a sync rule with a different suffix, and it gets the `.meta.json` case backwards. The Studio plugin already sends `ClassName` from the live DataModel, where `RunContext` is sitting right there, so that would cover more for anyone using it. Properly it probably wants an optional `runContext` in the sourcemap from Rojo's end.

1.69.0.

PR #1595, opened 2026-08-19
Respect .client.luau and .server.luau suffixes when classifying script context

dev-bap:
Addresses #1594.

`writePathsToMap` decides a script's context from className alone, so a `Script` with `RunContext = Client` reads as server and gets offered `ServerScriptService` modules. This checks the file suffix first when the class is a bare `Script`, and falls back to the class name when there isn't one:

```cpp
if (node->className == "LocalScript")
    node->scriptContext = ScriptContext::Client;
else if (node->className == "Script")
    node->scriptContext = scriptContextFromFilePathSuffix(node).value_or(ScriptContext::Server);
else
    node->scriptContext = parentNameContext;
```

`LocalScript` stays unconditionally Client. That one is a class invariant rather than a guess, and letting a filename override it could only make things worse. Anything that isn't a script still inherits from its container, as before.

Worth being upfront about what this leaves alone. Rojo lets a sync rule use whatever pattern it likes, so a project naming its files `*.rc-client.luau` is still misclassified. Same for `RunContext` set through a `.meta.json`, where a `.server.luau` declared as client reads as server, which is what happens today anyway. In all of those the class-name fallback applies, so they keep their current behaviour rather than getting a worse one. Covering the rest properly probably wants the run context in the sourcemap itself, which is Rojo's end.

Tests: four in `AutoImports` covering instance and string requires from both a client and a server RunContext script, and one in `Sourcemap` checking the classification directly, fallbacks included.

Full suite passes locally, including `--new-solver` and `--fflags=true`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Fixed `@self` string-require aliases resolving from the filesystem instead of the sourcemap tree for non-DataModel roots ([#1511](https://github.com/JohnnyMorganz/luau-lsp/issues/1511))
+- Auto-import server/client boundary filtering now respects `.client.luau`/`.server.luau` file suffixes, fixing misclassification of `RunContext = Client` scripts emitted by Rojo's `emitLegacyScripts: false` ([#1594](https://github.com/JohnnyMorganz/luau-lsp/issues/1594))
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


JohnnyMorganz:
Seems reasonable, but wondering if we'd be better off asking the Rojo folks if they could include the RunContext metadata in the sourcemap for us to use? That would help the custom sync rules too I guess?

dev-bap:
Agreed, that's the right place for it. A `runContext` field would cover the sync rule and `.meta.json` cases this PR can't. I've opened https://github.com/rojo-rbx/rojo/issues/1317.

JohnnyMorganz (approved):
Thanks!

PR #1598, opened 2026-08-22
Fix `@game` string requires in files outside the sourcemap

savruun:
Require-string autocomplete fails in the same files, through a separate path.

## Cause

Two independent code paths gate `@game` on the requirer having a sourcemap node.
Resolution goes through `resolveStringRequire`. Completion goes through
`RobloxStringRequireSuggester`. Each needs its own fix.

### Resolution

`RobloxPlatform::resolveStringRequire` returns early to the filesystem fallback
when `context->name` is not a virtual path. A file with no sourcemap node keeps
its filesystem module name, so the whole `@` block is skipped. The fallback
finds no `game` alias in `.luaurc` and treats `@game` as a directory name.

The `@game` branch reads the sourcemap root and the require string only. It
reads nothing from the requirer. `@game` is absolute: it is rooted at the
DataModel. The gate is too broad.

The instance-expression form is not gated. `RobloxPlatform::resolveModule`
returns `game` with no context check, so `game.ReplicatedStorage.Packages.fluid`
gives full type information in the same file where `require("@game/...")`
errors. The two spellings mean the same thing and behave differently.

### Autocomplete

`RobloxStringRequireSuggester::getNode` looks the requirer up in
`virtualPathsToSourceNodes` and returns a `FileRequireNode` on a miss. That node
offers filesystem children and `.luaurc` aliases only. It cannot list `@game` or
walk the DataModel, so the same file gets no suggestions inside a `@game/...`
string even though the require resolves once it is fully typed.

## Fix

Resolution: handle `@game` before the context gate, and delete the
now-unreachable branch below it.

Autocomplete: `SourcemapAwareFileRequireNode` subclasses `FileRequireNode`. It
keeps the filesystem behaviour and adds the built-in `@game` alias, which does
not depend on the requirer having a node. `FileRequireNode` members move from
private to protected so a platform can subclass it. Only the requirer node needs
this. Once the path descends into a directory, filesystem children are already
correct.

Three sites now match `@game`, so the alias parsing moves into a shared
`parseAliasRequire` helper.

Behaviour that does not change:

- A user-defined `game` alias in `.luaurc` still wins, for both resolution and
  the suggested alias list.
- `@self` and relative requires from a file outside the sourcemap still resolve
  on the filesystem. They need a filesystem base, so the fallback is correct.
- With no sourcemap loaded, `rootSourceNode` is null and nothing changes.

## Reproduction

```sh
mkdir -p repro/src repro/dist repro/packages && cd repro

cat > packages/fluid.luau <<'EOF'
local fluid = {}
function fluid.create(class: string) return class end
return fluid
EOF

cat > src/a.luau <<'EOF'
local fluid = require('@game/ReplicatedStorage/Packages/fluid')
return fluid.create
EOF
cp src/a.luau dist/a.luau

### DIFF
--- a/src/include/Platform/StringRequireSuggester.hpp
+++ b/src/include/Platform/StringRequireSuggester.hpp
@@ -29,7 +29,9 @@ public:
     std::vector<std::unique_ptr<Luau::RequireNode>> getChildren() const override;
     std::vector<Luau::RequireAlias> getAvailableAliases() const override;
 
-private:
+    // Protected rather than private so a platform can subclass this node and keep its filesystem
+    // behaviour while it adds platform-specific aliases.
+protected:
     Uri uri;
     bool isDirectory = false;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Fixed `@self` string-require aliases resolving from the filesystem instead of the sourcemap tree for non-DataModel roots ([#1511](https://github.com/JohnnyMorganz/luau-lsp/issues/1511))
+- Fixed `@game` string requires and their autocomplete failing in a file that the sourcemap does not cover, such as build output. `@game` is absolute, so it no longer needs the requiring file to have a sourcemap node
 
 ### Changed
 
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: src/include/Platform/StringRequireSuggester.hpp [hunk 1]
```diff
@@ -29,7 +29,9 @@ class FileRequireNode : public Luau::RequireNode
     std::vector<std::unique_ptr<Luau::RequireNode>> getChildren() const override;
     std::vector<Luau::RequireAlias> getAvailableAliases() const override;
 
-private:
+    // Protected rather than private so a platform can subclass this node and keep its filesystem
+    // behaviour while it adds platform-specific aliases.
```
JohnnyMorganz:
```suggestion
```

JohnnyMorganz (approved):
Thank you!

Issue #1250, opened 2025-11-09
First argument does not get skipped in tooltip for method calls on functions returned by type functions


StevenDahFish:
### Reproduction Code
```luau
type function repro()
	return types.newfunction({ head = { types.any, types.number } }, { head = { types.boolean } })
end

local foo: { bar: repro<> } = {
	bar = function(self: any, n: number): boolean
		return true
	end
}

foo.bar(foo, 2)
foo:bar(2)
```

Using the luau-lsp extension, the tooltip doesn't skip the first argument when writing out methods defined by a type function.

<img width="446" height="68" alt="Image" src="https://github.com/user-attachments/assets/1440f0ee-44cb-4891-89fc-762c8e942b53" />

In comparison to Roblox Studio, which properly omits the first argument.

<img width="340" height="90" alt="Image" src="https://github.com/user-attachments/assets/320474d4-1f2e-4249-9a1e-5cce2b7ec283" />

### Expected Behavior
(this is by removing the typing from `foo`)

<img width="530" height="63" alt="Image" src="https://github.com/user-attachments/assets/5e2db15e-dcff-4a67-8394-f5c90ad50a54" />

Note that the autocomplete does correctly only suggest that one argument is available while typing the method, as well as **not** having an "Argument count mismatch" error. It is only for the tooltip that doesn't skip the first argument properly.

JohnnyMorganz:
Unfortunately we can't get it to just show `function foo:bar(number): boolean` because the `bar(number): boolean` part is generated from Luau's `Luau::toStringNamedFunction`, and Luau will only skip it if `ftv.hasSelf` is true - https://github.com/luau-lang/luau/blob/master/Analysis/src/ToString.cpp#L1662-L1668

However, we can atleast get it to show similar to the expected behaviour you see.

PR #1613, opened 2026-08-30
Hide implicit self from __call signature help

### LINKED ISSUE #1597: __call showing self parameter when calling table
The self parameter thats autofilled in by __call is visible when calling the table
Im using the vscode extension

## how to reproduce:
1. create a table and give it a metatable with __call
2. type out the code that calls that table

## example:

```luau
local module = setmetatable({}, {__call = function(self,value)
    print(`printed out: {value}`)
end})

module()
```

Image

Companion:
## Problem

Signature help for a table invoked through its `__call` metamethod includes the first metamethod parameter:

```luau
local module = setmetatable({}, {
    __call = function(self, value: string)
    end,
})

module()
```

Luau supplies `self` implicitly, so callers cannot provide it. However, signature help currently displays:

```text
function module(self: any, value: string): ()
```

and reports both values as caller provided parameters.

## Change

Build `__call` signature help from a copy of the metamethod type with its implicit first argument removed. The corrected signature is:

```text
function module(value: string): ()
```

This removes `self` from both the displayed label and the LSP parameter list while preserving parameter offsets, active parameter tracking and documentation indices for the remaining arguments.

Ordinary function calls and existing method call (`:`) behaviour are unchanged. The separate method call label behaviour tracked by #1250 remains outside the scope of this change.

Fixes #1597.

## Testing

Added focused signature help tests covering:

* Removing the implicit argument from the label and parameter list
* `__call` metamethods containing only `self`
* Multiple explicit parameters and their label offsets
* Active parameter tracking
* Variadic parameters
* Parameter documentation indices
* Mixed named and unnamed arguments
* Unchanged ordinary function behaviour

The new tests were verified to be load bearing by temporarily disabling the trimming and documentation offset behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Fixed `@self` string-require aliases resolving from the filesystem instead of the sourcemap tree for non-DataModel roots ([#1511](https://github.com/JohnnyMorganz/luau-lsp/issues/1511))
+- Fixed Signature Help showing the implicit first argument of a `__call` metamethod, which is supplied by the table being called rather than by the caller ([#1597](https://github.com/JohnnyMorganz/luau-lsp/issues/1597))
 
 ### Changed
 
--- a/src/operations/SignatureHelp.cpp
+++ b/src/operations/SignatureHelp.cpp
@@ -119,7 +119,8 @@ std::optional<lsp::SignatureHelp> WorkspaceFolder::signatureHelp(
     std::optional<size_t> activeSignature = std::nullopt;
     std::vector<lsp::SignatureInformation> signatures{};
 
-    auto addSignature = [&](const Luau::TypeId& ty, const Luau::FunctionType* ftv, bool isOverloaded = false)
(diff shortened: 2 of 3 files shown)


JohnnyMorganz (commented):
I wonder if this could be simpler if we just build off the ` if (idx == 0 && candidate->self)` check. Instead, we can create a copy of the FunctionType and set `hasSelf` to true on it. Then we don't need to create this custom `trimmed` FunctionType.

Might need to do something on `toStringNamedFunction` to simplify it to hide / skip `self` parameter there

Companion:
> I wonder if this could be simpler if we just build off the ` if (idx == 0 && candidate->self)` check. Instead, we can create a copy of the FunctionType and set `hasSelf` to true on it. Then we don't need to create this custom `trimmed` FunctionType.
> 
> Might need to do something on `toStringNamedFunction` to simplify it to hide / skip `self` parameter there

That makes sense ill use a copied FunctionType with hasSelf set reuse the parameter skipping logic and update the signature formatter accordingly that should remove the custom trimming and documentation offset ill retain the regression coverage and check ordinary function and method behavior

JohnnyMorganz:
Thanks, unfortunately I already cut a release - can you move the changelog entry?

Companion:
Thanks! I've moved the changelog entry to Unreleased under Fixed.

